PR #59, opened 2020-06-11
Fix: Dragging crew members to new roles throws an error

mxvs:
Dragging previously added crew members to a new role after reloading or restarting the application (for example when starting a new gaming session) throws the following error and you can't change roles.

`starship.js:355 Uncaught TypeError: Cannot read property 'id' of null
 at ActorSheetStarfinderStarship._onDragCrewStart (starship.js:355)`

For some reason the Actor **_token_** is being referenced (which is not always available) when the `actorId `is already available in a constant in the same function. This small change fixes that bug so that reassigning crew members always works.

### DIFF
--- a/src/module/actor/sheet/starship.js
+++ b/src/module/actor/sheet/starship.js
@@ -340,7 +340,7 @@ export class ActorSheetStarfinderStarship extends ActorSheetStarfinder {
             data: actor.data
         };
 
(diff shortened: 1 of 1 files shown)


wildj79:
Nice catch! Gotta love weird edge cases ;) Merging.

PR #63, opened 2020-06-15
New JournalEntry Compendium to make more confortable our party

decad7never:
Add new JournalEntry compendium inside system.json
Add JournalEntry with dragndrop item for (All Weapons & Armors from the CRB / Spells for Technomancer and Mystic)
Armor augmentation is only the list today, we works on the display and dragndrop item.


wildj79:
This pull request looks good.

PR #72, opened 2020-06-23
Add a lots of weapons and armors inside equipments compedium.

decad7never:
update also list of weapons and armors inside the rules compendiums


decad7never:
update french translation & some fix

decad7never:
Fix the spellselector :)

wildj79:
I'm going to let it slide this time, but in the future I would _highly_ suggest that you create separate branches for each change you'd like to make and then create separate PR's for them as well.

PR #75, opened 2020-06-28
First step for improve action chat box

decad7never:
- Attack Roll (with damage / save)
- Display Range + Remaining Ammo

Same template for
- Attack Role
- Attribute Role
- Save Role
- Skills Role

no change for all other chatbox action.

Next step => remove manual roll without the custom template


wildj79:
After a cursory glance, this looks OK. But, since this is changing core functionality to a fair extent, I'm going to scrutinize the code a little more. I'm going to leave this one open for a bit until I can get a chance to go over the code changes a little more in depth. Feel free to continue to push changes to this branch until I can give it a proper code review.

decad7never:
Can give my the way to modify all other chatbox ? it seems more complicated to me, i have modify only roll that used the function d20Roll.

Else, it can be a first step to add on the next release before make the job for all rolls.

Thanks.

PR #77, opened 2020-06-29
Updated Short Rest interface

TimToxopeus:
I have updated the Short Rest interface:
- It is fully localized, relevant keys have been added or updated in the other language jsons, as requested by Logmelna.
- The short rest chat messages have been localized too.
- Restoring stamina is now a checkbox you can tick rather than a separate button that immediately performs the rest.
- Performing Rest is now done by clicking on the Rest button.


src/templates/apps/short-rest.html
```diff
@@ -1,9 +1,7 @@
...
-        <div class="form-fields">
-            <button id="spend-rp">Spend RP</button>
-        </div>
-    </div>
+    <p>{{localize 'STARFINDER.RestSDescription'}}</p>
+    <label class="checkbox">
+        <input type="checkbox" id="restoreStaminaCheckbox"> {{localize "STARFINDER.RestSActionHeader"}}
+    </label><br/>
 </form>
+<br/>
```
wildj79, line 7: Not sure that we need and explicit break here. If wee need more space at the bottom of the form it would be better to use `css` styling instead of `html` markup.

src/templates/apps/short-rest.html
```diff
@@ -1,9 +1,7 @@
...
-    <div class="form-group">
-        <label>Spend a Resolve Point</label>
-        <div class="form-fields">
-            <button id="spend-rp">Spend RP</button>
-        </div>
-    </div>
+    <p>{{localize 'STARFINDER.RestSDescription'}}</p>
+    <label class="checkbox">
+        <input type="checkbox" id="restoreStaminaCheckbox"> {{localize "STARFINDER.RestSActionHeader"}}
+    </label><br/>
```
wildj79, line 5: Not sure that we need and explicit break here. If wee need more space at the bottom of the form it would be better to use `css` styling instead of `html` markup.

src/templates/apps/short-rest.html
TimToxopeus: A good point. I have adjusted the margin on the rest form so that the checkbox has a little room to breathe.

TimToxopeus:
A good point. I have adjusted the margin on the rest form so that the checkbox has a little room to breathe.

PR #86, opened 2020-07-03
Counter management solarian fix sfrpg

decad7never:
Fix the Counter Classes Management, its broken since the last release.
And i fix also know issues - we can manage now multiple actor (player or npc) with the solarian feature.


decad7never:
I thnik its ok now. Can you confirm me please.

PR #90, opened 2020-07-04
Added Critical Damage to item sheets

TimToxopeus:
Critical damage was missing from the item sheets and roll implementation, so I have implemented the damage roll to the item sheet, hooked it up with the data fields, and integrated it with the damage roll when the user does a critical damage roll.


src/module/dice.js
```diff
@@ -149,6 +150,17 @@ export class DiceSFRPG {
                 let mult = 2;
                 roll.alter(add, mult);
                 flavor = `${title} (Critical)`;
+
+                if (criticalData != undefined) {
```
wildj79, line 154: The projects preference is to use the `!==` operator whenever possible rather than the dubious `!=`. Same goes for `===` vs `==`. Please correct.

wildj79:
Looks good. Merging.

PR #96, opened 2020-07-12
Added inline roll data support to rolling items/features

TimToxopeus:
When rolling an item or feature on a character sheet, it was not possible to make an inline roll in the description that included some data. For example, [[1d6 + @abilities.str.mod]]

I added actor data to the roll so the roll has proper context.


wildj79:
Cool. Easy enough fix.

PR #100, opened 2020-07-18
Updating spell ranges

Juxtaposedwords:
Changes are not all inclusive. Notable exclusions: anything with 'see text' or  the spell 'hold monster'.

Updating  ranges are formatted consistently (ie 'medium' is not listed as a unit, instead 'ft' is listed).  

Changes are all based off of aonsrd.com.  I've done a good bit of spot checking.  I am nevertheless too sensible of my defects not to think it probable that I may have committed many errors.


wildj79:
The ranges were done this way intentionally. I see what you're trying to do here, which is to keep the spell definitions in sync with the data from the books. That works great when you're playing at the table, but when you start considering automation from a VTT's standpoint, it's not so great. It makes more sense to mark these in terms of actual ranges so that module authors can calculate ranges easier. I appreciate the contribution, but the current system is working as intended.

Juxtaposedwords:
I'm a touch confused as I'm hearing you say that spell entries having different entries for the same range classes makes the system easier for VTT automation.  Could you help me understand how storing a range value differently for two spells helps with VTT? I would expect we would want all 'long', 'medium', and 'short' range spells to be written consistently.

Currently spell ranges are not consistently encoded.  Two long ranged spells:
* [Acid Rain](https://github.com/wildj79/foundryvtt-starfinder/blob/development/src/packs/spells.db#L20) uses 'ft' as a unit
* [Clairaudience/Clairvoyance](https://github.com/wildj79/foundryvtt-starfinder/blob/development/src/packs/spells.db#L28) uses 'long' as a unit

wildj79:
Ok, in that case Clairaudience is incorrect and it would seem I need to remove some options from the drop down. It’s my intention that the ranges be fully stated in feet and not use the short, medium, or long range short hand.

Juxtaposedwords:
For close, medium, long , and set ranges(ex: 130 ft) the proposed change would have always provided the unit 'ft'. Only in the other case of 'personal', 'touch', 'planteary', 'plane', "system-wide", and "unlimited" did I use a name in place of 'ft'. 

Would it be best for me to submit another change with which only updates the 'ft' unit changes (ie short, medium, long, and standard ft unit)? If that is the case, would you prefer 'touch' or 'personal' was 0 ft?

wildj79:
I'll reopen the PR, and I think 'touch' or 'personal' are fine.

Juxtaposedwords:
Per your request, the latest version should only have 'touch', 'personal', or 'ft' as the unit.

wildj79:
OK, looks good. Merging.

PR #101, opened 2020-07-19
Fixed overflow on ability boxes

TimToxopeus:
If the user enters an extremely large base attribute value, such as 123456, the ability score box would resize the input field out of the container, leaving the user unable to access the value. This fix prevents the text from overflowing out of the container.

I also cleaned up the CSS a little.


wildj79:
Looks good, merging.

PR #105, opened 2020-07-20
Added support for child modifiers

TimToxopeus:
Updated modifiers system for player characters:
* Added support for child modifiers from items.
* Added modifiers to the sheets of archetypes, augmentations, classes, equipment, feats, races, themes, and weapons.
* Added passing of child modifiers to closures.
* Removed Item and Misc categories from character modifiers sheet.


wildj79:
Ok, looks good. Merging.

PR #109, opened 2020-07-21
Force damage type

decad7never:
Add new damage type (Force) used by magic missile spells for example.


wildj79:
Just an FYI, if you are going to be creating multiple PR's, and you're creating separate branches for each, make sure that you always check out development before creating the next branch. It looks like you finished one set of changes, then created your new branch based on the previous branch, so all of your changes ended up in the last PR. I'm going to accept this one and close the other two, since this one has all of the code changes in it.

PR #110, opened 2020-07-24
Fix feat dynamic modificators

decad7never:
just a little fix about dynamic modificators for all feats


wildj79:
Looks good, merging.

PR #111, opened 2020-07-26
Drone Sheet

TimToxopeus:
Here it is, first submit for the new player character drone sheet! It's got a lot of content, so take a moment to review it.

It is a pretty thorough implementation of the drone mechanics, so I think I caught everything, but I probably made a mistake or two along the way as well. The main reason I am pushing this now is to enable cmd to start creating compendium items for the drone chassis and mods, I do not expect the data to change super majorly.

It's a fairly automated screen, but all of the automations are overridable through either config, or through modifying what the chassis and modules grant your drone. I am not fully finished with the modifiers integrations, but since this will likely change come 0.7 I didn't want to wait for modifiers.

I have also set up all the localizations, I think I caught everything, but it is always possible I missed something. I suppose Decad7never will let me know soon enough.

Finally, this pull request also includes a minor retouch to the inventory screen for player characters, notably the equip/unequip column has been moved into the actions, saving up a lot of screenspace for item names. It worked better for the drone sheet too, which was the main driver for this change.


wildj79:
Looks good. Thanks for the submission!

PR #113, opened 2020-07-28
Drone sheet improvements

TimToxopeus:
Fixed an error when creating a fresh drone actor and added a nice display when the drone has no skills available to it.


wildj79:
Looks good, merging.

PR #114, opened 2020-07-28
Fixed drone defense stat

TimToxopeus:
Forgot to include dexterity in the drone's AC calculations.


wildj79:
Looks good, merging.

PR #117, opened 2020-07-28
Ability scores not applied to saves.

TimToxopeus:
Con, Dex, and Wis scores weren't applied to saving throws.


wildj79:
Nice catch!

PR #119, opened 2020-07-28
Fix an error in drone saves

TimToxopeus:
There was a copy paste mistake in the calculation, my bad. :-)


wildj79:
🤦 I should have caught that before I merged the first request.

PR #124, opened 2020-07-31
Update spells.db

Juxtaposedwords:
Updating class, concentration, school, and source (with page numbers too).

For comparing: I downloaded each file, sorted the lines in each, and then used JSON prettify tool to break out the entries.  
Online diff: https://www.diffchecker.com/WUINmpax


wildj79:
Looks good. Merging.

PR #128, opened 2020-08-05
Drone Sheet Improvements

TimToxopeus:
* Added support for free drone modules, such as the initial chassis modules, or the Skill Unit (Ex) class feature.
* Added a label indicating the drone chassis, which turns red if no chassis is installed.
* Fixed the ability score tooltips.
* Default size is "Medium", not "med".


wildj79:
Looks good, merging.

PR #135, opened 2020-08-10
Translation fr

decad7never:
Drone translation
add missing translation key (drone) like unity for french translation
rework on fr translation from users feedbacks


TimToxopeus:
I have one suggestion, please move the unit in localization from SFRPG.DroneSheet.Speed into a separate category. Something like SFRPG.Units.Feet, SFRPG.Units.Metric, etc.

This will allow us to use units and unit conversions in other places than just the DroneSheet, even if it's only the drone sheet that supports it.

decad7never:
Its done i have made the modification with "SRFPG.Unity.Speed"

decad7never:
Move to SRFPG.Units.Speed

wildj79:
You missed changing it on `ja.json`, also, could you move SFRPG.Units.Speed to above the SFRPG.LocalizationTerminator field?

decad7never:
Its ok now. ja translation added, and i have change the position of the new translation key above the SFRPG.LocalizationTerminator field

PR #136, opened 2020-08-12
Update races.db

supernovajm:
Added races. Added descriptions from aonsrd


TimToxopeus:
Looks good, verified it, seems everything is a-ok! 👍

PR #139, opened 2020-08-14
Weapons and Armor added to equipment.db

supernovajm:
added stats for ~1000 weapons and armors, sourced from Jimbles' Starfinder Equipment Master List


TimToxopeus:
933 new json files, 933 new lines of entry in the equipment.db compendium, everything looks okay.

supernovajm:
Changes made to crit damage, and removed @bonus. I agree it would be nice to have the source page numbers, but the doc I was working off of didn't have them, so the time investment would have been exponentially higher.

PR #148, opened 2020-08-22
Ability Score now calculates Theme and Race ability mods

saffron-sd:
I added Theme and Race ability bonuses to Ability Score calculation

image


wildj79:
Me and @TimToxopeus have been discussing this PR some offline. We had some issues with it, but after looking a little closer, the issues have more to deal with inconsistencies on how the system is handling theme and race vs anything that you're code is doing. There is a concern that this might cause some confusion about where these values are coming from and I think this PR highlights some other issues that we are going to need to address at some point, like how do we handle ASI's and what not. But, I don't think that should prevent us from moving forward with a merge.

PR #152, opened 2020-08-23
Added ability check and ability checks modifiers

saffron-sd:
Ability Check and Ability Checks modifiers were only partially implemented so I finished them.

I'm not sure how to handle tooltip for these modifiers, Ability tooltip already displays Ability Score modifiers and ability check modifier wouldn't look good there


wildj79:
Code looks good, merging.

PR #158, opened 2020-08-24
Com

ShadowK13:
Added all COM classes and class features, including classes features for the CRB races. Added Witchwarper spell list with links. Minor change to Mystic and Technomancer spell lists to give the final "No Level" spell its own heading.


wildj79:
This looks good, merging.

PR #166, opened 2020-08-26
A aliens

Cryover:
Adds aliens starting with A based on https://www.aonsrd.com/Aliens.aspx?Letter=A

### DIFF
--- a/src/items/alien-archives/aeon_guard.json
+++ b/src/items/alien-archives/aeon_guard.json
@@ -1,0 +1,1533 @@
+{
+  "name": "Aeon Guard",
+  "type": "npc",
+  "data": {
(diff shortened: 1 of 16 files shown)


TimToxopeus:
I see a couple of non-A aliens, was that intentional?

Cryover:
> I see a couple of non-A aliens, was that intentional?

Yes, since i use the link in the last comment and they are inside the A category and are inside the "Family" Starting with A for Example:
- Aeon, Tekhoinos
- Angel, Barachius
- Azata, Tritidair
- Archon, Power Archon
- Akata, Void Zombie

TimToxopeus:
Gotcha :) Merging!

PR #171, opened 2020-09-01
Alien Archive Compendium - S Aliens

saffron-sd:
I added all available non-starship aliens starting with letter S (54 aliens), sourced from: https://www.aonsrd.com/Aliens.aspx


TimToxopeus:
Nice work :)

PR #173, opened 2020-09-04
Update text format for universal creature rules. (Lesson learned)

CptTwinkie:
My formatting depended on not so great header text. The improved text broke it as would be expected to happen. So here's the fix.


TimToxopeus:
Nice. 👍

PR #186, opened 2020-09-17
Added nonlethal as a damage type w/ localization

CptTwinkie:
I noticed this one missing from immunities but it was required by some NPC or monster I was importing.


CptTwinkie:
[[FEAT] Add nonlethal damage option to "Damage immunities"](https://trello.com/c/0YgUor7I/58-feat-add-nonlethal-damage-option-to-damage-immunities)

PR #188, opened 2020-09-25
Compendium update + Contributing.md changes

fmordashev:


CONTRIBUTING.md
```diff
@@ -33,11 +33,23 @@ The easiest workflow is to work through Foundry, and follow the following steps:
 2. Fill out the details, modifiers, etc.
 3. Right-clicking the item in the Items sidebar tab and select Export Data.
 4. Save the JSON file into the appropriate compendium subfolder.
+    * Make sure the name of the JSON follows these simple rules
```
CptTwinkie, line 36: I'm not sure the filename is the important part. It's the "Name" value inside the file which should be like the book. All of the json is renamed by cook, as far as I know.

fmordashev: Okay, had a different impression on how that works. Reflected the statements from above in another commit

fmordashev:
Added alien archive letters B and C. JMac gave me the jsons and asked to process them

TimToxopeus:
Thanks! :-)

CptTwinkie:
[Missing from equipment compendium](https://trello.com/c/edLiq41i/64-missing-from-equipment-compendium)

PR #191, opened 2020-10-04
German translation

jfastnacht:
Just in case someone else wants to work on that as well, you can also fork mine or send me a message to cooperate.


jfastnacht:
There are a lof ot Weapon Special Properties missing, since I only own the Starfinder Core Rulebook in German and most of them are from the Starfinder Armory. Therefore I cannot provide the missing translations right now.

According to https://www.aonsrd.com/WeaponProperties.aspx?ItemName=All the following books are needed:
* Alien Archive (Alien-Archiv https://www.ulisses-ebooks.de/product/236888/)
* Character Operations Manual (Starfinder - Einsatzhandbuch Charaktere https://www.ulisses-ebooks.de/product/319788/)
* Near Space (AFAIK not translated)
* Pact Worlds (Starfinder - Paktwelten https://www.ulisses-ebooks.de/product/266094/)
* Starfinder 3: Splintered Worlds (Starfinder - Tote Sonnen 3 - Splitterwelten https://www.ulisses-ebooks.de/product/247429/)
* Starfinder 5: The Thirteenth Gate (Starfinder - Tote Sonnen 5 - Das dreizehnte Tor https://www.ulisses-ebooks.de/product/256768/)
* Starfinder 15: Sun Divers (AFAIK not translated)
* Starfinder 19: Fate of the Fifth (AFAIK not translated)
* Starfinder 29: The Cradle Infestation (AFAIK not translated)
* Starfinder Armory (Starfinder - Rüstkammer https://www.ulisses-ebooks.de/product/266092/)

@wildj79 is there a general consensus on this? Literal translation until someone fixes it or rather keep the English translation until a proper source can be provided?

wildj79:
@jfastnacht Hmm... interesting question. I'm OK with just using literal translations until official translated text becomes available. Also, the next stable version of Foundry is due out in the next couple of days. If you can finish this translation before then, I would love to go ahead and merge this in for the next release. :)

jfastnacht:
Tbh, if there isn't any consensus, I'd rather stick to the English terms than confusing the GMs and players with made-up ones. In this case they can at least refer to the English descriptions when looking for the rules.

wildj79:
Fair enough. Everything looks good, so I’m gonna go ahead and merge it.

chief-wrench:
Hi everyone, is there interest in more localized german descriptions? I have already extracted lots of text from the official PDFs. I would be interested in adding german descriptions for items etc. 
Browsing through the translations for other languages it looks like nobody has translated item descriptions. Is that true? Is it even possible currently to translate more than the stuff already present in de.json?

wildj79:
Just a point of order: commenting on an already closed pull request is generally not a good thing. Comments like this are hard to find, and if it weren't for the fact that I get email notifications on stuff like this, it's very likely it would have gotten buried and never answered.

With that out of the way, translating compendium items isn't something that Foundry supports out of the box. There are modules that try to work around this, but it's pretty hard to do. One way to do this is to have separate compendium entries for each language you'd want to support. This is kind of out of scope for our system. Adding a separate compendium for each language would increase the size of the system fairly significantly, so we are not really interested in adding them. We are leaning heavily on the modding community to supply translation modules with the translated compendium items.

But, we would welcome another pull request with updated translation strings for the things that are easily translated.

jfastnacht:
@chief-wrench checkout https://gitlab.com/riccisi/foundryvtt-babele
You should start a separate repository with the translation, like https://gitlab.com/zenvy/foundryvtt-lang-de-pf1

chief-wrench:
@jfastnacht & @wildj79 Yes, since I wrote my comment I did some digging. Sorry for necro-ing a closed issue and CU in a new issue. Thanks for showing a newcomer the ropes...

PR #218, opened 2020-11-18
Patch 2

msbranin:
Updated themes that had the wrong info.


msbranin:
We stall have themes like athlete that Allow you to choose a +1 to str, dex or con... What would be the best way to handle those? Either the Ability boost needs to be modifiable or we could create 3 different Athlete themes, each with a different + 1ability boost.

src/items/themes/giantblood.json
```diff
@@ -11,7 +11,7 @@
     "type": "",
     "modifiers": [],
     "abilityMod": {
-      "ability": "str",
```
CptTwinkie, line 14: Maybe there's something I'm not understanding here but the old value looks correct to me
image

msbranin: Whoops That was a mistake. Can you look over merc I know it was comepletely wrong in the original file as well as priest had the wrong stat mod

msbranin: Wait. In Foundry when you look at the Giantblood theme is Says +1 Con which is where the initial mistake came from.
Giantblood

CptTwinkie (approved):

CptTwinkie:
This still needs a cook but the json seems to be correct now

TimToxopeus:
Looks fine, I'll merge it and cook it myself. Thanks!

PR #224, opened 2020-12-01
J, P-R spells

ThroughlyDruxy:
Minor corrections and some added spells. Corrected name from "Junk Sword" to "Junksword" to accurately reflect the name in book.


TimToxopeus:
Awesome!

PR #225, opened 2020-12-02
Fix modifiers off-target, off-kilter and flat-footed; can be added with charsheet checkbox.

Bagenholm:
Fix modifiers off-target, off-kilter and flat-footed; can be added with charsheet checkbox. 

Does not toggle flat-footed when off-kilter.


TimToxopeus:
The change looks okay, nice work. :)

Before I merge this, did you test what happens to existing characters that have, for example, checked the 'offkilter' setting before this change, and how their data is after this change? Does it break anything?

PR #227, opened 2020-12-05
MinorGuiFixes

CptTwinkie:
Fix for bold headers in item details
Match button shadows to their background colors


TimToxopeus:
Seems harmless enough, merging, thanks!

PR #230, opened 2020-12-11
Spells

ThroughlyDruxy:
Addition of spells from AA4 and some other corrections


TimToxopeus:
Awesome stuff. :) We've got ourselves a scoop now. ;)

PR #238, opened 2020-12-30
Shields update

CptTwinkie:
Includes latest errata...
- updated item stats 
- updated physical item traits for HP, AC, and Hardness
- equipped shields without proficiency do not provide a bonus (reflected in descriptions)
- added wielded bonus to AC values for PCs (0 if not proficient)
- added tooltip values and penalties for PCs (if not proficient)

image

### DIFF
--- a/src/items/equipment/dousing_shield_advanced.json
+++ b/src/items/equipment/dousing_shield_advanced.json
@@ -24,14 +24,14 @@
         "mod": ""
       },
       "hp": {
-        "value": 6,
-        "max": ""
+        "value": 51,
+        "max": "51"
       },
       "hardness": {
-        "value": ""
+        "value": "29"
       },
       "ac": {
-        "value": ""
+        "value": "10"
       }
     },
     "activation": {
--- a/src/items/equipment/dousing_shield_basic.json
+++ b/src/items/equipment/dousing_shield_basic.json
@@ -24,14 +24,14 @@
         "mod": ""
       },
       "hp": {
-        "value": 6,
-        "max": ""
+        "value": 21,
+        "max": "21"
       },
       "hardness": {
-        "value": ""
+        "value": "9"
       },
       "ac": {
-        "value": ""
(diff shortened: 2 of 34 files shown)


TimToxopeus (changes requested):
Missed localization opportunities:
When making changes, it is desirable to set up localization for those fields while you're there anyway.
item/item.js: Pushes some strings into the props without localization.
item/sheet.js: Pushes some strings into the props without localization.
calculate-base-armor-class.js: "unlimited" could be localized.

Code style nitpick:
In calculate-base-armor-class.js you have an "else {" on a separate line. Project code convention states that it should be "} else {"

Question:
In calculate-base-armor-class.js you create a tooltip with a negative shield profession bonus, but this negative score is never added as a shield bonus to the final score. Is this how it is supposed to work?

CptTwinkie:
> In calculate-base-armor-class.js you create a tooltip with a negative shield profession bonus, but this negative score is never added as a shield bonus to the final score. Is this how it is supposed to work?

Yes. The shield bonus when not proficient is 0 and is applied to the bonus. I calculated the negative value on the tooltip just to make it seem obvious to players that the bonus is being calculated but not applied. It's also consistent with the armor tooltips which I updated to make it clear they are for armor.

> Missed localization opportunities:
> When making changes, it is desirable to set up localization for those fields while you're there anyway.
> item/item.js: Pushes some strings into the props without localization.
> item/sheet.js: Pushes some strings into the props without localization.
> calculate-base-armor-class.js: "unlimited" could be localized.
> 
> Code style nitpick:
> In calculate-base-armor-class.js you have an "else {" on a separate line. Project code convention states that it should be "} else {"

I'll remove the offending newline and localize the file some more later today.

CptTwinkie:
image

TimToxopeus (approved):

TimToxopeus:
Looks good, merging it :)

PR #241, opened 2021-01-01
AC calculations now correctly handle multiple shields equipped at the same time

CptTwinkie:
In the tooltip below you see two shields with different names. "1st shield" has proficiency checked but "2nd shield" does not. They are otherwise identical.
image
image


CptTwinkie:
Also added ACP handling
image
image

TimToxopeus:
👍

PR #247, opened 2021-01-15
All the Starship weapons

ThroughlyDruxy:
Here it is peeps. All the weapons have been added (that I could find) from SOM, CRB, PW, NS, and various Alien Archives and Adventure Paths. There are bound to be an error here or there so if you notice them, let me know and I'll fix it.


TimToxopeus:
Awesome!

PR #248, opened 2021-01-16
Removed AA4 aliens with SF art 

ThroughlyDruxy:


TimToxopeus:
😢 👋 Bye aliens-with-external-art-links. May your return be swift and clean!

PR #258, opened 2021-01-26
Add the rest of the missing technological items

twrightsman:


twrightsman:
@msbranin I'm guessing we can close this pull request since it seems redundant to what you've done?

TimToxopeus:
Thanks Travis, looks like you're right, though, only some of the files conflict, so it seems there's still some items you had that weren't in Michael's commit? Could you take a look to make sure we're not just closing a good pull request too early?

msbranin:
I think the non conflicts are due to the names used, I used the names from the SRD. For instance you names your binders as binders but I named them Restraints, Binders as they are listed in the SRD

twrightsman:
That sounds reasonable @msbranin, I'll check the SRD next time. I was just reading from the CRB for these and the naming can be confusing. Thanks!

PR #266, opened 2021-01-31
Corrected Playable Races

ThroughlyDruxy:
Corrected all the playable races with mod support and correct sizes.


wildj79:
Looks good, merging.

PR #283, opened 2021-02-13
Added attacks to vehicles.

rrenna:
I've added basic Vehicle attacks to the Vehicle sheet. The Attacks Tab now contains an inventory of Attacks. This inventory can be filled with Weapons (mounted, can be used while driving) and Vehicle Attacks (a new item type).

**Weapons**
This is unpolished and not complete, as the majority of vehicles do not have separate attacks from their collision attack. This will also require the concept of passengers to be implemented, as the attack roll should be modified by the dex mod of a specific character in the vehicle. This will also require a new alert which asks the user if the vehicle is stopped, moving or moving full speed as each applies a different attack modifier.

**Vehicle Attacks**
Intended for the Collision attack that vehicles have, but could be used to represent other similar attacks (for instance if a vehicle had spikes that did Piercing damage that requires a Fort save). These have no attack and can only be rolled for damage. 

There's two primary uses for this in tactical combat: Ram and Run Over.

image

When ramming there's a reflex save to avoid (and damage is doubled):

image

When running over there's a reflect save to halve the damage:

image

**Next Steps**

I would like feedback on how i've implemented this. I don't know if Ram and Run Over should be a user option when asking for Damage instead of separate attacks (similar to Normal and Crititcal damage), the wrinkle is the save for each is different (halving damage vs. avoiding damage completely).

There's a few known issues:
- Weapons don't really work yet
- The new item type is listed as "vehicleAttack" when item types are listed in a drop down from the New Item button in the Items tab

### DIFF
--- a/src/lang/en.json
+++ b/src/lang/en.json
@@ -47,7 +47,8 @@
         "TypeChassis": "Drone Chassis",
         "TypeMod": "Drone Mod",
         "TypeShield": "Shield",
-        "TypeWeaponaccessory": "Weapon Accessory"
+        "TypeWeaponaccessory": "Weapon Accessory",
+        "TypeVehicleAttack": "Vehicle Attack"
     },
     "SFRPG": {
         "AbilityActivationTypesDay": "Day",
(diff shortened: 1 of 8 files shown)


src/lang/en.json
```diff
@@ -1053,6 +1055,12 @@
                 "AllowedArmorType"   : "Allowed Armor Type",
                 "Any"                : "Any"
             },
+            "VehicleAttack": {
```
rrenna, line 1058: I need to remove these keys, they are no longer used.

PR #284, opened 2021-02-13
Added attacks, passengers and piloting skill checks to vehicles

rrenna:
I've added basic Vehicle attacks to the Vehicle sheet. The Attacks Tab now contains an inventory of Attacks. This inventory can be filled with Weapons (mounted, can be used while driving) and Vehicle Attacks (a new item type).

**Weapons**
As the majority of vehicles do not have separate attacks from their collision attack. I've implemented the ability to add weapons to vehicles for reference but the user will have to roll the attack and damage themselves.

2021-02-13 at 8 11 PM

**Vehicle Attacks**
Intended for the Collision attack that vehicles have, but could be used to represent other similar attacks (for instance if a vehicle had spikes that did Piercing damage that requires a Fort save). These have no attack and can only be rolled for damage. 

There's two primary uses for this in tactical combat: Ram and Run Over.

image

When ramming there's a reflex save to avoid (and damage is doubled):

image

When running over there's a reflect save to halve the damage:

image

**Attack and Piloting mods**

I've added attack and piloting mod values to the vehicle details. The piloting mod is used to roll piloting checks from PC or NPC passengers, the attack values are currently only used for reference.

image

**Passengers**

I've add passengers (pilot, complement, passengers) to vehicles, with a UI that functions similar to starship crew.

passengers

complements

 Piloting checks can be rolled from this screen when a PC passenger list is being used.

image

When using NPC passengers, I've put a piloting check beside the piloting mod for convenience (there's no concept of which position is using the check so no need to force the user to switch tabs).

image

**Future Enhancements**

- Add an attack and damage button to weapons from PC passengers. This will require the attack roll be modified by the dex mod/BAB of a specific character in the vehicle. This will also require a new alert which asks the user if the vehicle is stopped, moving or moving full speed as each applies a different attack modifier.

- Add an attack and damage button to weapons from NPC passengers. This will also require the same alert referenced above which asks the user if the vehicle is stopped, moving or moving full speed as each applies a different attack modifier.

- Add vehicle subsystems, factor mods of subsystems into attack and piloting rolls from vehicle sheet.

- List total passenger capacity (and current count) along with if there's currently a pilot in header.


src/lang/en.json
```diff
@@ -1053,6 +1055,12 @@
                 "AllowedArmorType"   : "Allowed Armor Type",
                 "Any"                : "Any"
             },
+            "VehicleAttack": {
```
rrenna, line 1058: I need to remove these keys, they are no longer used.

TimToxopeus:
Took a quick look through the already submitted code. Stylistically it looks in line with the rest of our setup and I'm pretty happy with what I see so far. :-)

Couple of minor nitpicks:
- We try to keep localization file keys sorted alphabetically. It's a bit of a pain in the ass sometimes, but it helps find things.
- We prefer duplication of localization lines over re-use of specific localization from elsewhere. I saw at least one StarshipSheet key in your code. The problem with that is if the starship sheet were to change, it might accidentally end up affecting the vehicle sheet.
- Some rogue newlines in _prepareItems()

Keep it up! 👍

rrenna:
Thanks for the feedback.

I'll implement your changes, but as far as json sorting i'll use this tool right before merging (https://novicelab.org/jsonabc/). It'll cause a ton of changes, hence leaving it til the end to avoid conflicts in the interim.

src/module/actor/actor.js
```diff
@@ -467,12 +467,13 @@ export class ActorSFRPG extends Actor {
      * Prompt the user for input regarding Advantage/Disadvantage and any Situational Bonus
      * @param {string} skillId      The skill id (e.g. "ins")
      * @param {Object} options      Options which configure how the skill check is rolled
+     * @param {string} modified     A string which adds a specific modifier to the roll formula
      */
-    async rollSkill(skillId, options = {}) {
+    async rollSkill(skillId, options = {}, modified = null) {
```
rrenna, line 472: @TimToxopeus @wildj79 Can I get your input on this change, I feel like there should be an easier way to do this.

Problem: I need to roll a crew member's piloting, from the vehicle sheet, modified by a value (a piloting mod) stored in the vehicle. 

The way I do this is grab a reference to the crew's actor object, call `rollSkill()` but pass in a new parameter I added called `modified`. I feel like there's a key in the existing `options` I should be using instead, but couldn't find an example.

src/module/actor/actor.js
```diff
@@ -575,11 +576,15 @@ export class ActorSFRPG extends Actor {
...
     }
 
-    async rollSkillCheck(skillId, skill, options = {}) {
+    async rollSkillCheck(skillId, skill, options = {}, modified = null) {
         let parts = [];
         let data = this.getRollData();
 
         parts.push(`@skills.${skillId}.mod`);
+
+        if (modified != null && modified != undefined) {
```
rrenna, line 585: Don't know enough about JS patterns to know if the undefined check is necessary.

src/module/actor/sheet/vehicle.js
```diff
@@ -303,7 +299,7 @@ export class ActorSheetSFRPGVehicle extends ActorSheetSFRPG {
             const actorId = $(event.currentTarget).parents('.crew').data('actorId');
             const crew = game.actors.get(actorId);
 
-            crew.rollSkill("piloting", {event: event});
+            crew.rollSkill("piloting", {event: event}, this.actor.data.data.attributes.modifiers.piloting);
```
rrenna, line 304: See comment above, not sure if I can include this in the `options` object instead of increasing parameter count.

src/module/actor/actor.js
TimToxopeus: Why not roll it in the same way we roll them for the starship? Create a roll context, add the actor as a context, then use @contextidentifier.skills.pil.mod in the roll formula? i.e. "1d20 + @driver.skills.pil.mod + @vehicle.some.modifier"?

src/module/actor/actor.js
TimToxopeus: General JS shenanigan is that you should use === and !== instead of != and ==. You don't have to test for both null and undefined, in this case I would just put if (!modified) instead.

src/module/actor/sheet/base.js
```diff
@@ -225,6 +225,7 @@ export class ActorSheetSFRPG extends ActorSheet {
 
         // Condition toggling
         html.find('.conditions input[type="checkbox"]').change(this._onToggleConditions.bind(this));
+
```
TimToxopeus, line 228: Rogue newline is rogue

src/module/actor/sheet/vehicle.js
TimToxopeus: Yeah, I would prefer you to directly call the DiceSFRPG.rolld20 function here instead with the properties set up through the roll formula and roll context instead.

src/module/item/item.js
```diff
@@ -74,6 +74,11 @@ export class ItemSFRPG extends Item {
             labels.kac = data.armor.kac ? `${data.armor.kac} KAC` : "";
         }
 
+        // Vehicle Attacks
+        else if (itemData.type == "vehicleAttack") {
+            //
```
TimToxopeus, line 79: This is a placeholder for further implementation, yes? Please mark this as // TODO: <reason for placing to do>

src/module/item/item.js
```diff
@@ -974,6 +1028,44 @@ export class ItemSFRPG extends Item {
         });
     }
 
+    async _rollVehicleDamage({ event } = {}) {
+        const itemData = this.data.data;
+
+        if (!this.hasDamage) {
+            throw new Error("you may not make a Damage Roll with this item");
```
TimToxopeus, line 1035: Please replace this with ui.notifications.error() call with a nice localized error message.

src/module/item/sheet.js
```diff
@@ -233,6 +240,9 @@ export class ItemSheetSFRPG extends ItemSheet {
             let alignedBonus = item.data.proficient ? item.data.bonus.aligned : 0;
             props.push(game.i18n.format("SFRPG.Items.Shield.ShieldBonus", { wielded: wieldedBonus.signedString(), aligned: alignedBonus.signedString() }));
         }
+        else if (item.type == "vehicleAttack") {
```
TimToxopeus, line 243: === instead of ==

TimToxopeus:
Looks good now, nice work! :)

PR #295, opened 2021-03-02
Added the shatter weapon property (introduced in Fly Free or Die)

rrenna:
Added the `Shatter` weapon property.


TimToxopeus:
Missing de, jp, and fr localization file entries.

rrenna:
> Missing de, jp, and fr localization file entries.

Good catch!

PR #301, opened 2021-03-03
Added Shatter weapon property equipment from Fly Free or Die #01

rrenna:
Should be merged after `Shatter` property is merged, adding as draft for now.


rrenna:
Merge conflict fixed.

PR #305, opened 2021-03-06
Added basic text formatting check to Cook

rrenna:
Added some basic text formatting checks to `name`, `img`, `hp`, `sp` and `source`.

image


TimToxopeus:
I see you added warnings as a separate count. What is the benefit of having warnings?
I personally feel like any mistake is a mistake that needs fixing, so in my opinion there's no need to differentiate between breaking errors and minor errors.

*cue flashbacks to projects compiling with 4723 warnings, all #pragma disabled so people can see their errors*

TimToxopeus:
Implementation wise, it looks pretty exhaustive, nice work. :-) I did notice some inconsistencies in your indentations and newlines, could you make it match the rest of the code?

rrenna:
> I see you added warnings as a separate count. What is the benefit of having warnings?
> I personally feel like any mistake is a mistake that needs fixing, so in my opinion there's no need to differentiate between breaking errors and minor errors.
> 
> _cue flashbacks to projects compiling with 4723 warnings, all #pragma disabled so people can see their errors_

No particular reason one way or another, do we want to treat grammatical/formatting mistakes as breaking changes or not. I figured as I was introducing something that was going to complain like crazy, it was a bit presumptuous to claim there's "errors" (and thus you need to fix all 400 of them now!).

I'm good changing it back.

TimToxopeus:
My vote's in favour of claiming there are errors that need fixing. And then organizing getting them fixed. :D

rrenna:
Warnings removed.  👍

PR #306, opened 2021-03-07
Added vehicle compendium

rrenna:
*NOTE: Requires vehicle sheet PR to be merged first* 

Added a vehicle compendium with all vehicles from the core rulebook.


TimToxopeus:
👍

PR #307, opened 2021-03-08
Added vehicle,armor, spell & race validation. Added source field and condition link checking. Fixed bug with source validation.

rrenna:
- Fixed a bug with cook validation, there was a mistake in my regular expressions for source.
- Various source fields now validated
- Armor type, race size, alien size now validated
- Weapon multi-attacks now validated
- Scans most biographies / descriptions for references to a Starfinder condition that doesn't have a link to the compendium
- Added an alternate cook command which skips formatting checks (they may be slower on older computers and you may just want access to the cooked data) 
- Updated CONTRIBUTING.md to mention the alternate cook command.

Cook will now check


rrenna:
Removed 3 rogue newlines.

TimToxopeus:
What is this function x for? :D

PR #312, opened 2021-03-09
Added environment / location / organization fields to biography.

rrenna:
I've implemented environment and organization fields in the biography tab as a 3rd sub-section: ecology.

I've also renamed `Biology` tab to `About`.

image


rrenna:
Some interest has been mentioned of re-combining biography and ecology into one tab, this is outside the scope of this PR, but a CSS expert is free to come in and work some magic if they can get it to lay out correctly.

TimToxopeus:
Implementation looks clean, only one change I'd like to see. The way you added the ecology and organization to the details section means all characters now have ecology and organization, including player characters, starships, drones, and vehicles. I'd like that to be moved into the NPC definition.

These elements are all merged together, so it should be okay to just move it to the NPC definition.

rrenna:
Didn't I add an "if npc" to the tab bar? Let me check.

PR #316, opened 2021-03-12
Added all vehicles from Armory.

rrenna:
Added all vehicles from Armory.
Added new weapon type (Fire and Force) and new vehicle types, as needed for the new vehicles.


TimToxopeus:
Next time, please separate compendium commits and 'feature' commits such as localization. :)

PR #318, opened 2021-03-13
Adds tooltips to weapon properties everywhere but chat cards

rrenna:
This draft is a proposal of how we can add weapon property tooltips everywhere except chat cards. The architecture required significant changes to how we pass around item properties, it previous assumed the property was always a localized string (ie. `"one",` `"force",` etc) and I changed it to be an object instead (which contains a `name` and `tooltip` property).

This change would allow easy implementation of tooltips on all other item tags (proficiency, armor KAC, starship weapon properties, etc).

Screenshots:

image

image

image

If we haven't entered a tooltip yet for the property, no tooltip is displayed on hover:

image


rrenna:
@TimToxopeus looking for feedback on implementation before I go ahead and add all tooltips.

PR #321, opened 2021-03-14
Source fixes and Condition Links

ThroughlyDruxy:
Fixed source field for many Aliens and Equipment.

Also fixed broken links for conditions and fixed some modifiers for those conditions.


gulpfile.js
```diff
@@ -236,652 +236,320 @@ async function unpackPacks() {
 var cookErrorCount = 0;
 var cookAborted = false;
 var packErrors = {};
-async function cookPacksNoFormattingCheck() {
```
rrenna, line 239: I think you accidentally undid my changes in cook.

rrenna (changes requested):

gulpfile.js
ThroughlyDruxy: I believe I fixed it

PR #326, opened 2021-03-16
All vehicles from all APs

rrenna:
I've added all vehicles from all APs. I've also ran `cook`.


TimToxopeus:
There is a conflict. :(

PR #328, opened 2021-03-18
All remaining vehicles. Added 1 entry to rules compendium which was referenced by Water Bear.

rrenna:
The special ability of the Waterbear requires the reader to reference the Environmental protection section of the armor rules. I've added the armor rules in the rules compendium in this PR so as to add the dynamic link now without having to revise this vehicle "one day".


rrenna:
Fixed conflict. Re-ran `cook`.

PR #343, opened 2021-03-26
Added tons of links to conditions, planets, dieties, races, etc.

rrenna:
Added rules for defining effects and vehicle combat rules.

Added links in various items which reference mind-affecting, sense dependent and language dependent effect descriptors, also linked most references to a condition.


rrenna:
Converting to draft, there's too many changes to make without me double-checking no garbage is being checked in, will convert back to normal PR when i've skimmed over every change and double-check.

PR #344, opened 2021-03-28
Added basic alien browser

rrenna:
As per user request, I've added a (very basic) alien browser.

image

image

Sort by name or CR.
Filter by size.
CRs which are 1/3 or 1/2 have special display logic.

image


rrenna:
I've added a type filter.

src/module/packs/alien-archive-browser.js
```diff
@@ -0,0 +1,156 @@
...
+            console.log(`SFRPG | Compendium Browser | ${pack.metadata.label} - ${content.length} entries found`);
+
+            for (let item of content) {
+
+                item = item.data;
+                item.compendium = pack.collection;
+                // Used for sorting and displaying CR
+                item.data.cr = item.data.details.cr;
+                // 1/3 and 1/2 CR aliens have special strings used to describe their CR rather than using the float value
+                if (item.data.details.cr == (1/3)) {
```
wildj79, line 45: You should prefer `===` over `==`.

src/module/packs/alien-archive-browser.js
```diff
@@ -0,0 +1,156 @@
...
+                    item.data.crDisplay = "1/3";
+                }
+                else if (item.data.details.cr == (1/2)) {
```
wildj79, line 48: You should prefer `===` over `==`.

src/module/packs/alien-archive-browser.js
```diff
@@ -0,0 +1,156 @@
...
+        return sortingMethods;
+    }
+
+    _sortByCR(elementA, elementB) {
+        const aVal = parseFloat($(elementA).find('input[name=cr]').val());
+        const bVal = parseFloat($(elementB).find('input[name=cr]').val());
+        if (aVal < bVal) return -1;
+        if (aVal > bVal) return 1;
+
+        if (aVal == bVal) {
```
wildj79, line 80: You should prefer `===` over `==`.

wildj79 (changes requested):
As in a previous PR, just some little nitpicky issues. 

But, I'm not sure what's going on with `src/module/packs/item-browser.js`. It looks like it's just a reorganization and whitespace normalization of the code, but it's kind of hard to read. I would prefer if these types of refactors were done inside of there own PR's. If there were changes beyond code cleanup, I can't tell.

TimToxopeus:
> But, I'm not sure what's going on with `src/module/packs/item-browser.js`. It looks like it's just a reorganization and whitespace normalization of the code, but it's kind of hard to read. I would prefer if these types of refactors were done inside of there own PR's. If there were changes beyond code cleanup, I can't tell.

I believe there are also some edits to support returning actors instead of just items. I also prefer if it was done separate from newline changes because now it is hard to diff in github. I actually had to pull it in and kdiff3 it with whitespace disabled.

@rrenna and I had a brief chat about it, and I think we should rename item browser to entity browser, to properly reflect the fact it no longer just returns items.

TimToxopeus:
I'm going to request holding off on this one until after we have 0.8 working. Changes have to be made to the item browser for 0.8 compat, and since this isn't merged or processed still, I would wait with merging.

rrenna:
Sounds good. I also need to confirm the changes in the parent class was just tab space normalization. So I have that feedback to address before we can merge anyways.

wildj79:
I'm going to change this one to a draft for now. We'll have to revisit this one after v0.8.x hits stable.

TimToxopeus:
@rrenna Any chances you are possibly going to do anything here, we've been v0.8 stable for a while now, and there was some work still to be done here I think. Particularly with the refactoring of the way the document browser base class was put together, so it better reflects both items and actors.

PR #358, opened 2021-04-01
Add additional weapon accessory requirements

rrenna:
Some weapon accessories require "Rail weapons & Smallarms" or "Melee & Smallarms" those have been added to the available options in weapon accessory details.


TimToxopeus:
Before I merge, just a thought. Would it be a better idea to convert the setting into a checkboxable multi-select thing? Lest we end up creating half a dozen potential combinations over the course of more books?

rrenna:
I'll repeat my Discord comment, as it's been buried: `there hasn’t been any accessories since they were introduced in armoury - so until we have a hint the concept will be expanded I don’t think it’s worth a big refactor
As the current cases cover all combination
But yeah if tech revolution is full of accessories we should refactor I think, maybe at the same time as vehicle type and weapon damage type`

TimToxopeus:
Alright, sounds reasonable. :)

PR #366, opened 2021-04-06
Added iconic character

rrenna:


wildj79:
This one looks good. I haven't gone through each entry with a fine tooth comb, but I think it's good enough 😉 . Merging.

PR #375, opened 2021-04-22
Update Gulp workflow for Docker

trechriron:
Updated gulp.js and package.json with copy action vs. symlink. Docker hates symlinks. Also added a buildDock and cleanDock command for easier usage and avoid impacting the current workflow.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -996,6 +996,67 @@ async function linkUserData() {
 	}
 }
 
+
+/********************/
+/*	COPY USER DATA	*/
+/********************/
+
+/**
+ * Copy build to User Data folder (for Docker - doesn't like symlinks)
+ */
+async function copyUserData() {
+	const name = 'sfrpg';
+	const config = fs.readJSONSync('foundryconfig.json');
+	console.log("THIS IS SPARTA!");
+	let destDir;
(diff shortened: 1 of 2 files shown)


wildj79:
The PR looks fine, and I don't have any issues with merging this. But, I am curious as to its purpose? The gulp tasks, in general, aren't meant for mainstream consumption, there only meant to be used in development. Is there any particular reason you are setting this up for docker?

trechriron:
I use Docker because I don't trust Windows to emulate the production infrastructure. I wanted it to run on the same server core. My hope would be to eliminate dev vs prod wonkiness. :-)

Docker hates symlinks, so I needed actual fs.copy operations. I didn't want to mess with the current workflow.

This could be useful for new devs looking to use Docker. There are several implementations out there for Foundry.

rrenna:
Sounds interesting. Maybe if it's merged and works well you can amend the contributing documents with alternate setup instructions ?

trechriron:
I would be happy to! I can include info and my docker-compose file.

wildj79:
Ok, sounds good. Since this doesn't have any affect on our current development process, and it might be useful anyway, I'm going to go ahead and merge it.

PR #376, opened 2021-04-22
Add Short Description to Actor

trechriron:
Added the short description to actors. Tested on Chrome and Edge. Could not test on Firefox.

### DIFF
--- a/src/templates/actors/parts/actor-biography.html
+++ b/src/templates/actors/parts/actor-biography.html
@@ -38,6 +38,7 @@
 <div class="flexcol">
     <nav class="sheet-subnavigation biotabs" data-group="biographyTabs">
         <a class="item active{{#if data.details.biography.value}} highlight{{/if}}" data-tab="biography">{{localize "SFRPG.ActorSheet.Biography.TabBio"}}</a>
+        <a class="item {{#if data.description.short}} highlight{{/if}}" data-tab="descriptionShort">{{localize "SFRPG.DescriptionShort"}}</a>
         {{#if isNPC}}
         <a class="item active{{#if data.details.ecology.value}} highlight{{/if}}" data-tab="ecology">{{localize "SFRPG.ActorSheet.Biography.TabEcology"}}</a>
         {{/if}}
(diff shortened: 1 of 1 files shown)


wildj79:
I do have some concerns with this PR.

1) Your current changes won't do anything. It's not enough to add new fields in the handlebars templates. Any new values that need to be saved with the actor need to be added to the `template.json` file, so that foundry knows that these are valid fields and to save those values when the actor data is written to disk.
2) You're using a new translation string, but you haven't defined it in the language files. The key needs to be added to the language files located in `src/lang`.
3) I'm not really digging your choice for where the data will live. There isn't really a need to create a new `description` attribute on the actor. I'd prefer that it be saved in `data.details.biography.shortDescription` or something similar. So, in `template.json`, it would make more sense to add this to `Actor.templates.common.details.biography`. 
4) I also don't think there should be a new tab for this. All of this descriptive information would be better served under the "Biography" tab. 
5) I think this field is better served as a multi line text box, rather than a full blown editor field.

trechriron:
I just followed what was in Item already (my bad in template.json). Should be straightforward to fix this however.

Do you want Item and Actor short description implementations to match? My suggestion;

Template.json (Item)
Item.templates.itemDescription.description.short (I didn't see this before, sorry) (already present)

js (Item)
data.description.short (already present)

Template.json (Actor)
Actor.templates.details.biography.short (to match item)

js (Actor)
data.details.biography.short (to match item)

Then, make them a field (multi line, simple text) under description or biography vs. a seperate tab. This sound like a plan?

The DescriptionShort entry does exist in all language files so I can use that so both implementation match?

TimToxopeus:
Wow this has been woefully forgotten about, so, yes, biography.short for actors seems the appropriate place in template.json.

I'm not 100% sure what is the nicest way of handling short biography tab-wise however. But I can see adding a Short Biography header followed by a multiline text box (3 lines max) at the top, then followed by a header that says Full Biography, and the rest is the rich text editor as a visually pleasing integration.

Please make sure that the limited-character html sheet also properly reflects the shortened biography on the HTML page. This might also apply for NPCs, Drones, Ships, Vehicles, etc. Please double-check each actor type and test with different permission levels.

Items indeed already have integration for short descriptions.

PR #378, opened 2021-04-22
Update Manifest with Media Param

trechriron:
Added the manifest+ media/images folder with two images in the specified sizes for use in the Forge Bazaar.


wildj79:
I'm going to have to check on the images for this one. I need to confirm if they are covered by Paizo's community use policy or not. If not, we might need to source a different set of images.

rrenna:
The project definitely deserves a pretty face on Foundry Hub!

wildj79:
So, the community use policy does allow the use of the cover image. But, it's very specific that you not alter that image. 

```
Product covers may not be cropped, color adjusted, edited, distorted, or modified.
```

I'm gonna have to be a stickler on this point. I cannot accept this PR unless you correct the images to comply with Paizo's community use policy.

trechriron:
Gotcha, what are your feels? The full cover won't fit in the suggested dimensions. Want me to just name something snazzy with my royalty-free collection that is all space-fantasy?

TimToxopeus:
Can't you just center the product cover on top of an alpha-transparent background? Resizing the Paizo images is permitted, so you could scale it down to whatever fits within the desired resolution, and fill the rest with transparentness?

TimToxopeus:
Created some non-cropped images and added them, closing this PR.

PR #379, opened 2021-04-22
Translation to Portuguese (Brazilian)

lipefl:
A friend did the translation. Can you add it? Thanks!


TimToxopeus:
I cannot read Portuguese (Brazilian), but the json file at a cursory glance looks like it's in a good shape. I'll merge it, thanks! :-)

PR #385, opened 2021-05-07
Update HP in Kiirinta json

sturteva:
Updated src\items\races\kiirinta.json with correct HP value

Looks like when I was just checking out the gulpfile.js some spaces were added.


wildj79:
Looks good. Merging.

PR #392, opened 2021-05-26
Fix Altronus Weapon Focus

### LINKED ISSUE #391: Iconic character: Altronus error and feature request
In all 3 versions the character has Weapon Focus (advanced melee weapons) but the actual feat is set to apply that bonus to small arms.
And as a request, it would be useful for new players if the Solar weapon was set up as a custom weapon, so players can immediately use it without having to look up the current damage in the class feature.

bluedreamer:
Fixes #391
Added Solar Weapon to inventory tab for Altronus
Set default resolve points to max on other characters
Set token disposition to friendly on all character


TimToxopeus:
Looks okay. :)

PR #402, opened 2021-06-04
Weapon manufacturers integration

tmcfarland:
Added in new container type for weaponManufacturers and adjusted all weapons to include the ability to add a weapon manufacturer.  New type for weaponManufacturers added in including all weapon manufacturers that currently exist.  Weapon Manufacturers were based off fusions since the modify weapons in a similar way.  

Addition of Shattercloak and adaptive serum mk 1, 2, 3 into the compendium.  Adjustment for Magical items to be able to have damage properties in the template to allow quick rolling of damage items.  Adjustment of armor upgrades to be able to have damage properties in the template to support electrostatic field and quick damage rolls based on retribution damage from similar armor upgrades.

Template for weaponManufacturer container:
```
        {
          "type": "slot",
          "subtype": "weaponManufacturer",
          "amount": 1,
          "acceptsType": [
            "weaponManufacturer"
          ],
          "affectsEncumbrance": true,
          "weightProperty": ""
        }

```


tmcfarland:
Removing this Pull Request and breaking it into 2 separate PRs that will break out the Weapon Manufacturer changes to code and the actual data file migration to update weapons as a secondary update to make it clear what has been updated and done.

PR #405, opened 2021-06-09
Python scripts for compendium contributors

ThroughlyDruxy:
3 scripts to make it easier to clean jsons for addition to compendiums

comp_actor_cleaner.py and comp_equipment_cleaner.py iterates through files in its directory and strips the files of "_id", "sort", and "exportSource" fields.

comp_equipment_id_migrator is more complex. It basically replaces old compendium items with the new modified items while preserving the "_id" of the old file and removing "sort", and "exportSource" in the new file.

### DIFF
--- a/scripts/comp_actor_cleaner.py
+++ b/scripts/comp_actor_cleaner.py
@@ -1,0 +1,44 @@
+# Load the JSON module and use it to load your JSON file.                       
+# I'm assuming that the JSON file contains a list of objects.
+import os
+import json
+
+for files in os.listdir("."):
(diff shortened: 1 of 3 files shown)


wildj79:
Sorry it took so long, but these look good. It would be nice to have some documentation on how to use these scripts, but it's not a show stopper. I'll go ahead and merge this.

ThroughlyDruxy:
Yeah I'll throw a readme up there for them. I did a bit of commenting but
probably people using it won't read the comments in the code.

PR #415, opened 2021-06-25
chatMessage changes

ironmonk88:
I have a module that would like to roll ability, skills and saving throws.  Problem is I don't want the request to automatically open a Chat Message.  I've attempted some interested code to work around it, but DnD5e has a handy option of requesting that a dice roll doesn't open a ChatMessage, and I'd very kindly like to add it to Starfinder.

### DIFF
--- a/src/module/actor/actor.js
+++ b/src/module/actor/actor.js
@@ -547,6 +547,8 @@ export class ActorSFRPG extends Actor {
             title:  `Ability Check - ${label}`,
             flavor: null,
             speaker: ChatMessage.getSpeaker({ actor: this }),
+			chatMessage: options.chatMessage,
+			onClose: options.onClose,									
             dialogOptions: {
                 left: options.event ? options.event.clientX - 80 : null,
                 top: options.event ? options.event.clientY - 80 : null
@@ -582,6 +584,8 @@ export class ActorSFRPG extends Actor {
             title: `Save - ${label}`,
             flavor: null,
             speaker: ChatMessage.getSpeaker({ actor: this }),
+			chatMessage: options.chatMessage,
+			onClose: options.onClose,							
             dialogOptions: {
                 left: options.event ? options.event.clientX - 80 : null,
                 top: options.event ? options.event.clientY - 80 : null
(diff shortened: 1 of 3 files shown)


wildj79 (changes requested):
This seems reasonable and it shouldn't cause any issues. My one critic is that there something funky going on with white space in your PR. If you could clean that up a bit, I think we can move forward.

ironmonk88:
That was weird.  I think my text editor was doing something strange with tabs and spaces.  I tried to clean it up, but it looks the same on my machine so I'm not 100% I got the spacing correct.

Also, thank you very much!  This'll make coding my module a whole lot easier.

TimToxopeus:
A couple of things from my side:
- You did not cover the scenario where the user is using the custom chat cards. If you go to system settings and enable "Use custom Starfinder chat card?", it uses a slightly different code path in dice.js. Please double-check these cases, in the code you can recognize them as SFRPGCustomChatMessage.
- Consider renaming chatMessage to showChatMessage, to emphasize it is not the actual chat message object, but instead a toggle for showing or hiding the chat message.

Some whitespace/coding convention nitpicks :)
- dice.js line 93: No { } around the if statement.
- dice.js line 121: Missing whitespaces after if and before {.
- dice.js line 429: Missing whitespaces after if and before {.
- item.js line 1136: You are using tabs instead of spaces.
- item.js line 1205: You are using tabs instead of spaces.

ironmonk88:
SFRPGCustomChatMessage should be corrected now.
I used `let useCustomCard = game.settings.get("sfrpg", "useCustomChatCards") && chatMessage;` so that it would just bypass the custom card altogether and go to the default if the chat message wasn't to be shown.

Fixed the white space issues.

I agree that chatMessage isn't as descriptive... but it is also used by other systems.  I just wasn't sure about maintaining consistency.  And... just because other systems use it, doesn't mean it's right.  Let me know what your preference is, and I can change it or leave it.  :-)

TimToxopeus:
My personal preference is to make things nice over how the others do it, but I don't know if it causes problems with modules. @wildj79 what do you think?

wildj79:
Sorry for the late reply. My personal preference is to err on the side of being too verbose and making things clear, rather than not verbose enough. But, I think this is fine for now. I'll leave this one open for a few more days if you want to go ahead and make the change, otherwise I think this one is fine to merge.

wildj79 (approved):

TimToxopeus:
I was going through pull requests and wanted to merge this one, but noticed there are merge conflicts that have originated in the months that followed the "few more days" 😅. Any chance to get these fixed?

Dosendusche:
created a new PR that just resolves the merge conflicts: https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/pull/845

LebombJames:
Closed in favor of #845

PR #423, opened 2021-07-08
Archetype Compendium Entries

LebombJames:
Added entries for all 30 archetypes to a new compendium.

They're currently reference-only, but I could add some modifiers in the future.

Passed cook with 0 errors, so fingers crossed everything is in order!


wildj79:
Looks good, merging.

PR #449, opened 2021-10-01
Fix for NPC spell attacks

J-Dawe:
This fixes an issue where NPC spell attacks were erroneously adding their Intelligence modifier when they shouldn't be.


TimToxopeus:
Hi, I like your change, but the code for it already exists, the if is just in the wrong order from the looks of it. If you look at line 644 in your case it already tests for npc type. I think we only need one of those checks, rather than two. :)

J-Dawe:
Done.

TimToxopeus:
Thanks 👍

PR #460, opened 2021-11-15
Implemented token hud with string layout and fixed click behavior

CptTwinkie:
[Added] Formatting on the TokenHUD. The name labels are optional, the removal all button is always there.
image

[Fixed] Click on the token statuses updates the actor sheets properly

[Change] The data format of the condition images to provide metadata required for modules. The effects of this can be seen in the final HTML
image


CptTwinkie:
I found a case where it was possible to sometimes get double icons on the token. I discovered I missed an await on the setCondition call which is why it was working so well in the first place. Once I added the await I got double icons consistently. After a little experimentation, I decided to allow the actor to control the token toggling since that is more reliable than a hybrid system. The one remaining bug has to do with conditions that toggle other conditions. There seem to be double updates happening there. I'll have to get into the code that handles those cases to find out why.

CptTwinkie:
The Flat-Footed check always had a double update. Now there's a triple update thanks to Foundry. There's a check that's maybe a little shortsighted in the code that checks if the effect data has an icon property. If it does then it takes an adventure through active effects and then calls toggleEffect again with more processed data that allows the icon to work properly in overlay mode. 
**To summarize**: I fixed everything but there is a triple update on the tokenHUD every time a condition changes and there's not much I can do about it unless foundry improves the conditions in their Token.toggleEffect method.

src/module/actor/mixins/actor-conditions.js
```diff
@@ -1,6 +1,12 @@
 import { SFRPG } from "../../config.js"
 
 export const ActorConditionsMixin = (superclass) => class extends superclass {
+    /**
+     * Check if the Actor has the condition.
+     * @param {String} conditionName The name of the condition. Must match any key from config.js SFRPG.statusEffectIconMapping. Case insensitive.
```
TimToxopeus, line 6: The conditionName is suggested to be case insensitive in your comments, but it is case sensitive.

src/module/actor/mixins/actor-conditions.js
```diff
@@ -11,6 +17,12 @@ export const ActorConditionsMixin = (superclass) => class extends superclass {
         return (conditionItem !== undefined);
     }
 
+    /**
+     * Get a condition Item from the actor.
+     * @param {String} conditionName The name of the condition. Must match any key from config.js SFRPG.statusEffectIconMapping. Case insensitive.
```
TimToxopeus, line 22: The conditionName is suggested to be case insensitive in your comments, but it is case sensitive.

src/module/actor/mixins/actor-conditions.js
```diff
@@ -21,21 +33,38 @@ export const ActorConditionsMixin = (superclass) => class extends superclass {
         return conditionItem;
     }
 
-    async setCondition(conditionName, enabled) {
+    /**
+     * Updates the Actor's conditions. Either adds or removes a condition Item as necessary to match the enabled argument.
+     * @param {String} conditionName The name of the condition. Must match any key from config.js SFRPG.statusEffectIconMapping. Case insensitive.
```
TimToxopeus, line 38: The conditionName is suggested to be case insensitive in your comments, but it is case sensitive.

src/module/actor/mixins/actor-conditions.js
```diff
@@ -21,21 +33,38 @@ export const ActorConditionsMixin = (superclass) => class extends superclass {
...
+     * @returns {Promise<*>} The Promise resulting from the create or delete Embedded Document call.
+     */
+    async setCondition(conditionName, enabled, { overlay = false } = {}) {
         if (!SFRPG.statusEffectIconMapping[conditionName]) {
             ui.notifications.warn(`Trying to set condition ${conditionName} on actor ${this.name} but the condition is not valid. See CONFIG.SFRPG.statusEffectIconMapping for all valid conditions.`);
             return;
         }
 
-        const conditionItem = this.getCondition(conditionName);
+        // Try to get status effect object as a workaround for a poorly conceived check in foundry.js Token.toggleEffect(...)
```
TimToxopeus, line 49: Status effect for loop can be replaced with something like:
let statusEffect = SFPRG.statusEffects.find(x => x.id === conditionName);

src/module/actor/mixins/actor-conditions.js
```diff
@@ -21,21 +33,38 @@ export const ActorConditionsMixin = (superclass) => class extends superclass {
...
+        let statusEffect;
+        for (const effect of SFRPG.statusEffects) {
+            if (effect.id === conditionName) {
+                statusEffect = effect;
+            }
+        }
+        
+        statusEffect = statusEffect ?? SFRPG.statusEffectIconMapping[conditionName];
```
TimToxopeus, line 57: Under what condition can statusEffect be null while statusEffectIconMapping isn't? This line seems unnecessary and prevents statusEffect from being const.

src/module/actor/mixins/actor-conditions.js
```diff
@@ -50,11 +79,8 @@ export const ActorConditionsMixin = (superclass) => class extends superclass {
                         const promise = this.createEmbeddedDocuments("Item", [itemData]);
                         promise.then((createdItems) => {
                             if (createdItems && createdItems.length > 0) {
-                                const updateData = {};
-                                updateData[`data.conditions.${conditionName}`] = true;
-                                this.update(updateData).then(() => {
-                                    Hooks.callAll("onActorSetCondition", {actor: this, item: createdItems[0], conditionName: conditionName, enabled: enabled});
-                                });
+                                this._updateActor(conditionName, true);
+                                Hooks.callAll("onActorSetCondition", {actor: this, item: createdItems[0], conditionName: conditionName, enabled: enabled});
```
TimToxopeus, line 83: Why did you take the onActorSetCondition callback out of the actor update? It was done on purpose so that the callback would not fire until the actor was fully finished updating.

src/module/actor/mixins/actor-conditions.js
```diff
@@ -66,14 +92,50 @@ export const ActorConditionsMixin = (superclass) => class extends superclass {
             if (conditionItem) {
                 const promise = this.deleteEmbeddedDocuments("Item", [conditionItem.id]);
                 promise.then(() => {
-                    const updateData = {};
-                    updateData[`data.conditions.${conditionName}`] = false;
-                    this.update(updateData).then(() => {
-                        Hooks.callAll("onActorSetCondition", {actor: this, item: conditionItem, conditionName: conditionName, enabled: enabled});
-                    });
+                    this._updateActor(conditionName, false);
+                    Hooks.callAll("onActorSetCondition", {actor: this, item: conditionItem, conditionName: conditionName, enabled: enabled});
```
TimToxopeus, line 96: Why did you take the onActorSetCondition callback out of the actor update? It was done on purpose so that the callback would not fire until the actor was fully finished updating.

src/module/actor/mixins/actor-conditions.js
```diff
@@ -66,14 +92,50 @@ export const ActorConditionsMixin = (superclass) => class extends superclass {
...
     }
+
+    /**
+     * Updates the actor data with the condition settings and then checks if Flat-Footed needs to be updated.
+     *
+     * @param {String} conditionName The name of the condition matching keys from config.js SFRPG.statusEffectIconMapping
+     * @param {Boolean} enabled If this value is true it enables the condition
+     * @private
+     * */
+    _updateActor(conditionName, enabled) {
```
TimToxopeus, line 111: _updateActor can be made async, and return the promise with the .then concatenation, to allow the callback to be performed after _updateActor finishes.

src/module/actor/mixins/actor-conditions.js
```diff
@@ -66,14 +92,50 @@ export const ActorConditionsMixin = (superclass) => class extends superclass {
...
+        const updateData = {};
+        updateData[`data.conditions.${conditionName}`] = enabled;
+
+        this.update(updateData).then(() => {
+            this._checkFlatFooted(conditionName, enabled);
+        });
+    }
+
+    /**
+     *
```
TimToxopeus, line 121: _checkFlatFooted has comment but without description.

src/module/config.js
```diff
@@ -1047,43 +1047,183 @@ SFRPG.CR_EXP_LEVELS = [
...
+        id: "stunned",
+        label: "SFRPG.ConditionsStunned",
+        icon: "systems/sfrpg/icons/conditions/stunned.png"
+    },
+    {
+        id: "unconscious",
+        label: "SFRPG.ConditionsUnconscious",
+        icon: "systems/sfrpg/icons/conditions/unconscious.png"
+    }
+]
```
TimToxopeus, line 1226: Does not end with ;

src/module/config.js
```diff
@@ -1267,6 +1407,8 @@ SFRPG.conditions = {
     }
 };
 
+SFRPG.conditionsCausingFlatFooted = ["blinded", "cowering", "off-kilter", "pinned", "stunned"]
```
TimToxopeus, line 1410: Does not end with ;

src/module/config.js
```diff
@@ -1047,43 +1047,183 @@ SFRPG.CR_EXP_LEVELS = [
...
-    "systems/sfrpg/icons/conditions/pinned.png",
-    "systems/sfrpg/icons/conditions/prone.png",
-    "systems/sfrpg/icons/conditions/shaken.png",
-    "systems/sfrpg/icons/conditions/sickened.png",
-    "systems/sfrpg/icons/conditions/staggered.png",
-    "systems/sfrpg/icons/conditions/stable.png",
-    "systems/sfrpg/icons/conditions/stunned.png",
-    "systems/sfrpg/icons/conditions/unconscious.png"
-];
+SFRPG.statusEffects = [
```
TimToxopeus, line 1050: With your changes to config.js, it might actually be worthwhile refactoring statusEffectIconMapping away entirely, and replace the checks inside the hasCondition, getCondition, and setCondition functions with a test against statusEffects.

src/module/token/token-hud.js
```diff
@@ -0,0 +1,124 @@
+export class SFRPGTokenHUD extends TokenHUD {
```
TimToxopeus, line 1: Please make a const pass on the entire file.

src/module/token/token-hud.js
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * @override
+     * Calls setCondition from the SFRPG actor before calling super
+     * @param {Event} event 
+     * @param {any} param1 - contains overlay boolean
+     * @returns super
```
TimToxopeus, line 8: Looks like it returns a boolean to me, and super is not called anywhere?

src/module/token/token-hud.js
```diff
@@ -0,0 +1,124 @@
...
+        }
+    }
+
+    /**
+     * @override
+     * Localizes and sorts the effects list before calling super
+     */
+    _getStatusEffectChoices() {
+        CONFIG.statusEffects = CONFIG.statusEffects.sort(function (a, b) {
+            let aid = (a.label != undefined ? game.i18n.localize(a.label) : a.id);
```
TimToxopeus, line 47: Localization of config elements should happen in sfrpg.js, check the hook for setup.

src/module/token/token-hud.js
```diff
@@ -0,0 +1,124 @@
...
+            let bid = (b.label != undefined ? game.i18n.localize(b.label) : b.id);
+            return (aid > bid ? 1 : (aid < bid ? -1 : 0));
```
TimToxopeus, line 49: Please look at String.prototype.localeCompare():
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

src/module/token/token-hud.js
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * @listens
+     * Sets all conditions to false
+     * @param {Object} event - standard event interface
+     */
+    static async onRemoveAllConditions(event) {
+        const statuses = this._getStatusEffectChoices();
+
+        for (const [k, status] of Object.entries(statuses)) {
```
TimToxopeus, line 116: I think there is merit to adding a getActiveConditions to actor-conditions.js that does something like this, just returns an array of ConditionItems currently present on the actor. I'm not sure how _getStatusEffectChoices operates in the parent class.

TimToxopeus (changes requested):
Did a pass through your work and left some comments and suggestions. Looking good so far!

src/module/actor/mixins/actor-conditions.js
CptTwinkie: Yes, you are correct. Even though there are comparisons made in lower case only there are also checks that are sensitive. I'll change the documentation.

src/module/actor/mixins/actor-conditions.js
CptTwinkie: Same as above

src/module/actor/mixins/actor-conditions.js
CptTwinkie: Same as above

src/module/actor/mixins/actor-conditions.js
CptTwinkie: Mistakes happen. You built the actor conditions using only the mappings but the token effects require the use of status effects. Since I'm essentially gluing those two together to make this work, I felt it was best to handle cases where someone did something bad like adding or removing a condition to only one list. Should I make the system less tolerant so that a local variable can be a const?

src/module/actor/mixins/actor-conditions.js
CptTwinkie: I think it got lost in the shuffle. You'll notice the update is calling a different method entirely. I'll make it a callback again.

src/module/actor/mixins/actor-conditions.js
CptTwinkie: same as above

src/module/actor/mixins/actor-conditions.js
CptTwinkie: duh!

src/module/token/token-hud.js
CptTwinkie: Yep, refactored out to fix some bullshit

src/module/token/token-hud.js
CptTwinkie: my bad. i didn't write this sort.

src/module/token/token-hud.js
CptTwinkie: fine. done. i'm not sure what the purpose of this pain in the ass is for but it's there.

CptTwinkie:
@TimToxopeus your comments should have been addressed by these commits

src/module/actor/mixins/actor-conditions.js
```diff
@@ -1,41 +1,88 @@
...
+     */
+    async setCondition(conditionName, enabled, { overlay = false } = {}) {
+        if (!this._isStatusEffect(conditionName)) {
+            ui.notifications.warn(`Trying to set condition ${conditionName} on actor ${this.name} but the condition is not valid. See CONFIG.SFRPG.statusEffects for all valid conditions.`);
             return;
         }
 
-        const conditionItem = this.getCondition(conditionName);
+        // Try to get status effect object as a workaround for a poorly conceived check in foundry.js Token.toggleEffect(...)
+        let statusEffect = SFRPG.statusEffects.find(effect => effect.id === conditionName);
```
TimToxopeus, line 75: const instead of let please. I don't think statusEffect is altered anywhere inside the function.

src/module/actor/mixins/actor-conditions.js
```diff
@@ -66,14 +111,51 @@ export const ActorConditionsMixin = (superclass) => class extends superclass {
...
+     *
+     * @param {String} conditionName The name of the condition matching keys from config.js SFRPG.statusEffects
+     * @param {Boolean} enabled If this value is true it enables the condition
+     * @private
+     * */
+    async _updateActor(conditionName, enabled) {
+        const updateData = {};
+        updateData[`data.conditions.${conditionName}`] = enabled;
+
+        this.update(updateData).then(() => {
```
TimToxopeus, line 135: This needs to be returned, return this.update ..

src/module/actor/mixins/actor-conditions.js
```diff
@@ -66,14 +111,51 @@ export const ActorConditionsMixin = (superclass) => class extends superclass {
...
+        });
+    }
+
+    /**
+     * Checks if the Flat-Footed condition should be enabled based on other conditions in SFRPG.conditionsCausingFlatFooted.
+     * @param conditionName The condition being added or removed
+     * @param enabled True if the condition is being added
+     * @private
+     */
+    _checkFlatFooted(conditionName, enabled) {
```
TimToxopeus, line 146: This should be marked async as well, and return this.setCondition .., promise should always be propagated up.

This way we can also ensure the onActorSetCondition callback is called after flat footed is potentially also applied.

TimToxopeus (changes requested):
Three final comments, almost there. :)

TimToxopeus (approved):
Looks good now.

PR #463, opened 2021-11-21
Remove attack and damage from Confused

CptTwinkie:
Pretty self-explanatory. For some reason, the Confused condition had an attack and damage. I have removed them.


TimToxopeus:
Very confusing indeed. :D

PR #470, opened 2021-12-03
Refactor global attack modifier to a getter

Bagenholm:
Refactoring the additional attack modifiers to a getter, which would allow monkey-patching it so a module could let you edit the modifiers without making any other changes to rollAttack().


TimToxopeus:
There is a bug when you open the dialog multiple times, it will keep appending to the config object. Localization is also missing.

However, as wildj79 is planning to ship a build today I will merge it in this case and fix it for you, as we discussed in Discord.

PR #481, opened 2021-12-21
Small tweak for the activation button on active feats

CptTwinkie:
When you have a feat with an activation cost of none the label on the button is set to null leaving a thin and difficult-to-click line instead of a properly sized button. This PR sets the label to "Activate" to indicate that the feat can be activated while not indicating a cost as would normally be the case


CptTwinkie:
image

PR #497, opened 2022-01-06
Data entry Dump

LebombJames:
This merge contains several (alternate) class features, spells, feats from TR, GEM, AA3, and a few APs, as well as fixing broken Weapon Spec links in classes. I haven't added all of the class features to their respective lists, but those could do with a full redo anyway at some point; most of them have lots of dead links.


wildj79:
Looks good. Merging.

PR #572, opened 2022-03-19
Fix CSS image paths with route prefix

### LINKED ISSUE #560: Missing icons with routePrefix
For example mousing over items/spells/etc. make the icon vanish while it's trying to render `/icons/svg/d20-black.svg`.

The paths should not start with slash to make them compatible with routePrefix usage.

For reference, routePrefix is mentioned here: <https://foundryvtt.com/article/configuration/>

mkahvi:
Without this some icons are missing entirely. Especially when hovering over features and such.

Fixes #560

### DIFF
--- a/src/less/actors.less
+++ b/src/less/actors.less
@@ -618,11 +618,11 @@
         }
 
         &:hover .item-image {
-          background-image: url("/icons/svg/d20-grey.svg") !important;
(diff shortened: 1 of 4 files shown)


TimToxopeus:
Thanks for fixing the bug! :D

PR #573, opened 2022-03-19
Add spellcasting ability selector to classes

LebombJames:
Currently, spell DCs always default's to your key ability score. This merge incorporates the dropdown into the calculation, allowing Precogs to have an easy method of setting their spellcasting, instead of manually changing every spell. The priority is:

1.  The ability set in the spell
2.  The ability set in the spellcasting dropdown
3.  The class's key ability score

Also set the default save DC of feats to half level rather than full level as that is usually the default for class features.

### DIFF
--- a/src/lang/de.json
+++ b/src/lang/de.json
@@ -521,6 +521,7 @@
         "ClassSkills": "Klassenfertigkeiten",
         "ClassWeaponProf": "Umgang mit Waffen",
         "ClassWillSaveProgression": "Progression der Willerettungswürfe",
+        "ClassSpellcastingAbility": "Spellcasting Ability",
         "Classes": {
             "IsCasterClass": "Handelt es sich bei der Klasse um einen Zauberwirker?"
         },
--- a/src/lang/en.json
+++ b/src/lang/en.json
@@ -521,7 +521,7 @@
         "ClassSkills": "Class Skills",
         "ClassWeaponProf": "Weapon Proficiencies",
         "ClassWillSaveProgression": "Will Save Progression",
-		"ClassSpellcastingAbility": "Spellcasting Ability",
+        "ClassSpellcastingAbility": "Spellcasting Ability",
         "Classes": {
             "IsCasterClass": "Is this a spell casting class?"
         },
--- a/src/lang/en.json
+++ b/src/lang/en.json
@@ -521,6 +521,7 @@
         "ClassSkills": "Class Skills",
         "ClassWeaponProf": "Weapon Proficiencies",
         "ClassWillSaveProgression": "Will Save Progression",
+		"ClassSpellcastingAbility": "Spellcasting Ability",
         "Classes": {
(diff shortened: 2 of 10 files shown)


TimToxopeus:
Seems to me the spell save DC should first try to pull the spellcasting ability score from the spellcasting class first, then use the item override, then use the spellbook override, then use the actor key ability score. Right now classes just have one ability score dropdown, do you want to take a crack at adding this feature?

LebombJames:
Ok, I've added a separate spellcasting ability to the class item. Spells will default to this over your KAS and the spellbook dropdown.

I think the priority should be as follows, however:
1. Ability in a spell
2. Class spellcasting ability
3. Spellbook ability dropdown
4. Key ability score (which is still set to the old KAS dropdown, not the one from classes btw. Shouldn't be too hard to change with the changes I've made)

This way, you can use the ability in the spell to setup multiclassing and spell-like abilities correctly.

Also haven't setup localisation yet: todo.

TimToxopeus:
Couple of nitpicks :)
1. Please change the tabs to spaces, let's keep the codebase consistent :D
2. Can you update the other localization files with the new keys as well?
3. In calculate-save-dc.js please rename SpellbookSpellAbility to spellbookSpellAbility, variables always start lowercase.

More serious issues however:
1. The selection of classSpellAbility seems to operate on the assumption there is a technomancer, mystic, witchwarper, or precog class, however, the classes array might be empty, so accessing it with [0] in that case would be illegal.
2. There are ways to cast spells without using cha as a non-spellcaster class. For example, feats that allow you to cast spells as if you are a specific class, or if you are a mechanic with the mechanic trick that grants you access to cast spells through spell chips. I'm not 100% sure what the proper solution here would be, but probably one would use the mechanic class' spellcasting ability score in that scenario. Spells have to select which class they belong to, perhaps we have to make that list show all classes rather than just the spellcasting classes, and then when you are searching the classes array, look for the class the spell is set to belong to.

LebombJames:
Ok, I've solved 1-3, and made classSpellAbility not throw an error if the class array is empty. 

I'm not really sure how to approach the final point. I had an idea of setting spellAbility in calculate-classes.js to default to the spellcasting ability, and then defer to the KAS if the class isn't a caster. Then, all classes would be passed into calculate-save-dc.js and used in classSpellAbility. I'd then create a separate preparation mode for Spell-like abilities that would force the ability to Cha.

But on the other hand, you could just use the spellbook dropdown if you've taken a spellcasting feat or you're a mechanic with spell chips, or the individual overrides if you're already a spellcaster. 

I also considered a class selection dropdown on spells, which would select the right ability score. Additionally, this would be useful on feats, so DCs that scale off class levels are correct.

Overall... 🤷‍♂️

TimToxopeus:
Very well, it is a bit of a complicated situation anyway. No need to withhold these fixes until we figure out how to tackle a very rare edge case.

Issue #458, opened 2021-11-14
Feature Request: add gunnery skill to npc crew on starships


Dosendusche:
Since i couldn't even find an acceptable workaround it would be nice to have this. It's strange nobody seems to have an issue with that.
Nearly all Paizo Statblocks for Starships have this skill.

CptTwinkie:
The workaround is piloting ranks. I personally don't like this as a workaround since it's a little confusing and it means you have a somewhat random number for piloting ranks. So I would also like to see gunnery listed as a skill. I would be perfectly happy to see it listed everywhere.

Although it could really be house-ruled to be used anywhere you are operating very heavy weaponry or perhaps call it remote weaponry such as aiming satellite lasers or mechanized planetary defense guns or something crazy like that.

Dosendusche:
i'm a little late sorry! 
Was added here: https://github.com/wildj79/foundryvtt-starfinder/releases/tag/v0.19.0
PR is mentioned above.
closing now, big thanks to @LebombJames.

PR #584, opened 2022-04-06
Implement Gunnery skill for NPC Starships

LebombJames:
I've seen that setting up NPC gunners is a frequent issue for some users on the discord, so I cooked this up to alleviate it. Just wanna gauge feedback for now, see if this is appropriate for the system. I was also thinking the calculation could be the max of the modifier and ranks, in order to not break existing NPC ships.


TimToxopeus:
I'm not sure this is correct behavior. Piloting mod is for piloting checks only, piloting ranks or BAB is for attacks. But allowing piloting mod you create a situation where the wrong value could be used. A better way would be to figure out how to rename the piloting ranks field into "Gunnery bonus" for NPCs specifically.

LebombJames:
image
Ok, got the ranks column to show "Gunnery Modifier" for only gunners. The issue with this however is that the column heading is incorrect if the gunner has any other skills.

image
I also started working on adding "gunnery" as a selectable skill and got it half-working, but it still has the issue of the column heading being incorrect. This method would still require an if statement on the `parts` declaration to use the gunnery "skill" modifier for NPC ship attacks. (I suppose you could use the modifier column for the bonus, as the language is slightly more intuitive)

We've got three different implementations here all with different pros and cons, but I think the pseudo-gunnery skill is the best.

LebombJames:
image
image

Got a gunnery skill working, left max(piloting ranks/gunnery mod) in for backwards compatability

LebombJames:
Wrote a migration, will set gunnery mod equal to piloting ranks, then remove the piloting entry.

LebombJames:
Aha, there's an issue for this. Closes #458

TimToxopeus:
I have nothing big to complain about, great work! :)

PR #587, opened 2022-04-08
Scaling Cantrips variant rule

LebombJames:
Feeling productive this week! 😁

Added a system setting to enable scaling cantrips. Upon enabling, it'll edit the affected spells in the spells compendium.

I was going to add a check to startup to ensure the setting is being followed, but I was concerned about init times so I left that out, let me know if I should pursue that.

The formula used is admittedly a really ugly lookupRange one, I'll see if I can make a non-lookupRange one, but it'll still be a bit ugly. At least when expanded, a non-lookupRange formula won't show *all* the dice.

Also I just put the function in item.js, let me know if there's a better place for it.

### DIFF
--- a/src/items/spells/energy_ray.json
+++ b/src/items/spells/energy_ray.json
@@ -93,6 +93,10 @@
       "dc": "",
       "descriptor": "negate"
     },
+    "scaling": {
+      "d3": true,
+      "d6": false
+    },
     "school": "con",
     "source": "CRB pg. 353",
     "sr": true,
--- a/src/items/spells/hazard.json
+++ b/src/items/spells/hazard.json
@@ -92,6 +92,10 @@
       "dc": "",
       "descriptor": "negate"
     },
+    "scaling": {
+      "d3": true,
+      "d6": false
+    },
     "school": "evo",
     "source": "COM pg. 136",
     "sr": true,
--- a/src/items/spells/injury_echo.json
+++ b/src/items/spells/injury_echo.json
@@ -94,6 +94,10 @@
       "dc": "",
       "descriptor": "negate"
     },
+    "scaling": {
+      "d3": false,
+      "d6": true
+    },
     "school": "trs",
     "source": "GM pg. 76",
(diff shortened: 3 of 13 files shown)


LebombJames:
Updated the formulas to be smaller, and not display every single dice used in it. Thanks to Iankid for help with this!

TimToxopeus:
A nice idea, I have some changes I'd like to request.
1. The supported spells are currently hardcoded, as per the spells variable. This is not very nice. I'd like to see spells eligble for scaling to have a setting, so that we can apply it without having to do name comparisons.
2. You are making local changes to the compendium. While not intrinsically bad, I would rather have a solution where we update the actors' spell items instead.
3. After switching to updating actors instead of the compendium, you can do bulk update on an actor for all items in 1 update, rather than 1 update per item. This greatly improves the performance cost.
4. The "scalingCantrips" function name is not a proper function name. A proper function name should always include a verb. In this case "onScalingCantripsSettingChanged" would be a good function name. The "on" nomenclature indicates it is an event that responds to the setting "scaling cantrips" that is changing.
5. Variables that do not get changed should be marked const, instead of let.
6. Our code convention within the project uses semicolons at the end of each instruction.
7. There are some weird indentations according to the github preview. Please make sure the indentations are neat and correct. If this is a fault in the github preview you can ignore this. :)
8. There are some double newlines in some places.

LebombJames:
Ok, re-implemented this feature with the requested changes.

- The affected spells now have a `scaling` object on them, that stores two booleans, one for if they use the d3 formula, and one for the d6 formula.
- The formulas of affected spells are automatically updated when dropped onto an actor from a compendium or the sidebar.
- Upon changing the setting, a "migration" is performed on all actors to alter the formulas of affected spells if they have any.

TimToxopeus:
I would suggest assuming CL as CR if that makes your life easier, but I don't know the NPCs that well to know all the exceptions that might apply.. Either way, great work, gonna merge this one in. :)

PR #589, opened 2022-04-11
Updated spellcasting classes with new spellcasting ability entries + a few finishing touches.

LebombJames:
Mystic, Technomancer, Witchwarper and Precog have had their spellcasting ability set correctly in accordance with the changes from #573.
Added a blank entry to the top of spellcasting ability so non-casters have no spell-ability, as opposed to all defaulting to strength. (This also means there's no reason to filter classes in calculate-save-dc.js anymore, all classes are passed in)
Hooked up the class KAS to spell DC calculation, will be deferred to if the sheet KAS selector has no value.
Spellbook ability has been given priority over spellcasting ability.


LebombJames:
Just noticed bonus spells per day are still calculating off of KAS, let me go fix that

TimToxopeus:
Looks okay now, is this ready for merge or are you still tinkering?

TimToxopeus:
Also, I recommend you don't commit .db files anymore; I usually do this after I merge a PR to ensure everything is okay. They cause conflicts. Run cook locally to verify no errors, then don't submit the .db files. :)

LebombJames:
Ok, noted on the dB files. Yep, this is ready to merge

LebombJames:
Resolved merge conflicts, still worth a recook just in case I messed something up 😛

TimToxopeus:
Mergin' :D

PR #591, opened 2022-04-14
Fix applying damage to starships throwing an error

LebombJames:
Currently applying damage to a starship throws an error, saying ChoiceDialog is undefined. Adding an import fixes that and it behaves as expected.


wildj79:
Thanks for this one. I'm going to see if I can release a hot fix for this one, since doing damage in starship combat is basically broken without this 😬 .

PR #597, opened 2022-05-03
Updated levels of mk 2 and 3 Culinary Synthesizer

LebombJames:
Quick fix; spotted the levels were all 1 on AoN, and the same was the case in the system.


wildj79:
This one looks pretty standard. Thanks for the fix on this one. Merging.

PR #598, opened 2022-05-06
Skill notes

LebombJames:
Example:
image

image

I used flavor to implement this rather than the same method as attack's roll notes and damage, just because I couldn't figure out how that worked. Don't think it'll be a major issue in the implementation though.


TimToxopeus:
Cute feature, I like it. 👍

PR #600, opened 2022-05-06
Mass migrated NPCs to have action targets

LebombJames:
May be a few outliers incorrectly changed (did my best to look for and change any), but those can be changed when discovered. It's better to have the majority correct (even then, the outliers will say what they *actually* are in the rules text)


TimToxopeus:
Agreed, this one is also a nightmare to verify so I'mma just merge it.

PR #605, opened 2022-05-08
Added NPC support to scaling cantrips

LebombJames:
The functions now give NPCs a formula based off of CR (because I could find no examples of an NPC with a CR different to their CL.) plus a general clean up.


wildj79:
Looks good, merging.

PR #610, opened 2022-05-09
Fix "undefined" on damage chat card if damage section has no name

LebombJames:
Currently, chat cards display an "undefined" on weapons with one damage section if the damage section has no name. This fixes that, and also removes the stray ":" caused by the same circumstances.


wildj79:
Looks good, merging.

PR #611, opened 2022-05-09
Fix popout styling

mkahvi:
This applies minor fix to make popout styling match that of the chat cards.


wildj79:
I think we'd prefer the alternative in this case.

mkahvi:
> I think we'd prefer the alternative in this case.

I have changed it to the alt selectors.

wildj79:
Looks good, merging.

PR #613, opened 2022-05-10
Auto add unarmed strike to new PCs

LebombJames:
This change adds the unarmed strike item to all new PCs to indicate that it's an option for them. (Also added the proper description to Unarmed Strike).
I thought about making this an optional system setting instead, thoughts?


wildj79 (changes requested):
I think the changes look good, but as you said, it probably makes sense for this to be a system setting.

LebombJames:
Sorry to complicate things, but I was working on this just as you requested those changes 😁

The above commits move the auto-linking and auto-unarmed strike to a new _preCreate method on the Actor class. This improves performance as (if my understanding is correct) the above changes are done *with* the creation of the new actor, rather than the new actor being created, then multiple entirely separate updates being ran in addition. I've also reduced it down to just one update push for both functions.

Made it a system setting as well.

PR #614, opened 2022-05-10
Added descriptors to spells and feats

LebombJames:
This'll be a fun one for data entry!
Added selectors for [effect descriptors](https://www.aonsrd.com/Rules.aspx?ID=158) similar to weapon properties for feats and spells, and display them in the item sidebar

image


src/module/item/sheet.js
```diff
@@ -256,12 +256,19 @@ export class ItemSheetSFRPG extends ItemSheet {
             )
         );
         } else if (item.type === "spell") {
+            const desc = (Object.entries(itemData.descriptors)).filter(e => e[1] === true).map(e => ({
```
wildj79, line 259: The open parentheses after the `=>` is unnecessary here. Just need the curly bracket.

src/module/item/sheet.js
```diff
@@ -256,12 +256,19 @@ export class ItemSheetSFRPG extends ItemSheet {
...
+                    name: CONFIG.SFRPG.descriptors[e[0]],
+                    tooltip: (CONFIG.SFRPG.descriptorsTooltips[e[0]]) ? CONFIG.SFRPG.descriptorsTooltips[e[0]] : null
+                })
```
wildj79, line 262: Ending parentheses is unnecessary here.

src/module/item/sheet.js
```diff
@@ -256,12 +256,19 @@ export class ItemSheetSFRPG extends ItemSheet {
...
+            )
```
wildj79, line 263: Missing semicolon after this parentheses.

src/module/item/sheet.js
```diff
@@ -273,10 +280,16 @@ export class ItemSheetSFRPG extends ItemSheet {
                 tooltip: null
             });
         } else if (item.type === "feat") {
-            props.push({
-                name: labels.featType,
-                tooltip: null
-            });
+            const desc = (Object.entries(itemData.descriptors)).filter(e => e[1] === true).map(e => ({
```
wildj79, line 283: same as above, the open parentheses aren't needed after the `=>`

src/module/item/sheet.js
```diff
@@ -273,10 +280,16 @@ export class ItemSheetSFRPG extends ItemSheet {
...
+                    name: CONFIG.SFRPG.descriptors[e[0]],
+                    tooltip: (CONFIG.SFRPG.descriptorsTooltips[e[0]]) ? CONFIG.SFRPG.descriptorsTooltips[e[0]] : null
+                })
```
wildj79, line 286: Don't need the ending parentheses.

src/module/item/sheet.js
```diff
@@ -273,10 +280,16 @@ export class ItemSheetSFRPG extends ItemSheet {
...
+            )
```
wildj79, line 287: Missing semicolon after parentheses.

src/templates/items/parts/item-descriptors.html
```diff
@@ -0,0 +1,9 @@
+            {{!-- Descriptors Template --}}
+            <div class="form-group stacked weapon-properties">
```
wildj79, line 2: It makes more sense semantically to use a custom class here instead of reusing `weapon-properties`. Maybe something like `item-descriptors`. Then in `src/less/item.less`, you can update this rule like this:

```less
/* ----------------------------------------- */
/*  Item Actions                             */
/* ----------------------------------------- */
.item-descriptors label.checkbox,
.weapon-properties label.checkbox {
  flex: 0 0 130px;
}
```

wildj79 (changes requested):

src/module/item/sheet.js
LebombJames: I noticed this when writing this code. It throws an error when those brackets aren't there. You'll see that the weapon property code on line 252 also has those brackets (that's where I copied the code from in the first place)

wildj79: Ok, I see what its doing here. It looks funky, but its working they way it should, so never mind ;)

wildj79 (approved):

wildj79:
Mergin'.

PR #615, opened 2022-05-11
Damage rolls below 1 return 1 non-lethal damage

LebombJames:
As per [CRB pg. 240](https://aonsrd.com/Rules.aspx?ID=103), damage below 1 becomes 1 point of non-lethal damage. 
(This may rarely come up, but it may come up a bit more after by archaic -5 damage PR 😛)


wildj79:
Looks good, merging.

PR #616, opened 2022-05-12
Migrate container item jsons as to not trip migration

LebombJames:
At the moment, a container migration trips if a container has any of the properties listed on line 955 of `actor-inventory-utils.js`, however, the container items in the compendium have some of those very properties, so I've removed them. 

The migration itself is still a bit problematic, as it doesn't work correctly if the items only have _some_ of the properties (see above 😉), and I may try to rectify that, but at least by changing the compendium JSONs the problem is no longer propagating.

### DIFF
--- a/src/items/equipment/backpack_consumer.json
+++ b/src/items/equipment/backpack_consumer.json
@@ -4,17 +4,6 @@
   "type": "container",
   "data": {
     "type": "",
-    "acceptedItemTypes": {
-      "augmentation": true,
-      "consumable": true,
-      "container": true,
-      "equipment": true,
-      "fusion": true,
-      "goods": true,
-      "technological": true,
-      "upgrade": true,
-      "weapon": true
-    },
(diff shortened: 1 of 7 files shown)


LebombJames:
Ok, at this point:

- Container items in the compendium have been migrated
- When a migration is performed on a container with *some* old values, if it has new values, those will be retained.
- However, the migration seems to be triggering every world boot because it doesn't remove the old values... and I can't figure out why it doesn't.

wildj79 (changes requested):
The `json` changes look good, but I'm not sure what you're trying to do in `actor-inventory-utils.js`. I think it might be better in this case to remove those changes from this PR and issue a new one for them, that way we can get the compendium changes in a release in the next day or two.

LebombJames:
Got it, will do.

wildj79:
Mergin'

PR #617, opened 2022-05-12
Add Starship Critical Damage Table

LebombJames:
How did this not exist already? 🤣
Simple merge, just adds the starship crit effect table as a rolltable.


wildj79:
Seems strait forward enough, merging.

PR #619, opened 2022-05-14
GM Mystic features added

Iankid:
Added GM Mystic features
Links in several Mystic connections fixed


wildj79:
Looks good, merging.

PR #624, opened 2022-05-17
Allow actor modifiers to be used in DR

### LINKED ISSUE #595: Enhanced Resistance not correctly parsing formulaic modifiers
Hello!

When attempting to apply Enhanced resistance to a character, any attempt to use a formula/reference to any value on the character sheet is unsuccessful. The calculation isn't made, and instead of displaying a value displays a blank space. When you take the specified damage type, it zeros out your HP/SP. Flat values still work.

I tested the following:

Energy resistance (Cold):
Single type of damage (cold) = zeroed out my HP
Multiple types of damage (including cold) = zeroed out my HP
Any type of damage (not cold) = worked just fine.

Damage Reduction:
Same results, but done with Piercing instead of Cold damage.

The following formula was used:

image

LebombJames:
Currently a value such as `@details.level.value` doesn't work with DR (and ER I suppose). Closes #595.

### DIFF
--- a/src/module/rules/actions/actor/calculate-damage-mitigation.js
+++ b/src/module/rules/actions/actor/calculate-damage-mitigation.js
@@ -36,7 +36,7 @@ export default function (engine) {
 
         for (const drModifier of damageReductionModifiers) {
             // TODO: Resolve formula; use RollTree, as it can complete synchronously
-            const resolvedModifierValue = tryResolveModifier(drModifier.modifier, rollContext);
+            const resolvedModifierValue = tryResolveModifier(drModifier.max, rollContext);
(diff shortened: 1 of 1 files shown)


Iankid:
Tested PR, Modifier Calculations now work for DR/ER. However, updates to sheet that affect the calculation do not update for Damage Reduction, unless you open the modifier then save it.

TimToxopeus:
I've looked at it, and I'm not sure the change is correct. It makes no sense it should resolve with .max instead of .modifier, .max should only be used for determining the bigger potential for stacking bonus evaluation.

I would like a little more details why it should use .max instead.

PR #630, opened 2022-05-23
Fix background repeating issues with certain class icons, plus wrapping long race text

LebombJames:
Before:
image

After:
image


wildj79:
Another simple fix that looks like it's working pretty good locally. Mergin'

PR #631, opened 2022-05-23
Tidy up scripts folder

LebombJames:
Sorted the scripts a bit instead of just throwing them into the scripts folder loose.

Also added a new js-ified item cleaner script that removes flags and ID and such. Haven't removed the old python scripts yet, but will do once I make an updated ID migrator script. Was there any reason actors and equipment were done in two separate python scripts?


wildj79:
Merging.

PR #633, opened 2022-05-26
ECM starship weapons use Computers ranks + Int to hit

LebombJames:
NPCs still use the gunner's attack bonus as outlined [here](https://www.aonsrd.com/Rules.aspx?ID=551)


TimToxopeus (changes requested):
I am not a fan of nested ternary operators, please refactor this so that we do not have nested ternary operators.

PR #634, opened 2022-05-27
Shift clicking a roll with the skip dialog option enabled will show the dialog window

### LINKED ISSUE #635: Checking Use Quick roll by Default doesn't allow shift+click to access the modifiers pop up window
Current Behavior:
check Use Quick Roll by default?
shift+click on an a roll
roll will happen in the chat window

Expected Behavior:
Check Use Quick Roll by default?
shift+click on a roll and open a pop up window to modify it
click roll on window

Attached a screenshot of the popup window for clarity's sake. If you need any additional info on reproduction steps let me know.
CleanShot 2022-05-26 at 19 54 03

LebombJames:
Currently, when the skip dialog option is checked, shift-clicking a roll has no effect. It would be more intuitive if this setting instead acted as an inversion of the default behavior: showing the dialog on a shift-click. 

Closes #635

### DIFF
--- a/src/lang/en.json
+++ b/src/lang/en.json
@@ -1987,7 +1987,7 @@
                 "Name": "Use custom Starfinder chat card?"
             },
             "UseQuickRollAsDefault": {
-                "Hint": "When clicking abilities that roll dice, the roll will act as if the Shift key was depressed by default. This will trigger a fast-forwarded roll.",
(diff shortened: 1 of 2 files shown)


wildj79:
This seems simple and strait forward enough. I ran some tests locally with it and seems to work fine, so merging.

PR #646, opened 2022-06-01
DC ammo types

### LINKED ISSUE #644: New Ammo types in DC
DC pg. 51
-Mood Goo
-Thasphalt
-Thasteron pellets

LebombJames:
Closes #644


TimToxopeus:
I don't see anything wrong here, but I think we need to start re-thinking the data entry approach, config file is starting to get bloated. 🤔 If more books add more ammo types, things'll get bloated sooner or later.

PR #650, opened 2022-06-02
Show message when applying critical damage (i.e from a nat 20) to a starship

LebombJames:
While there are messages for when critical thresholds are crossed, there is currently no message informing the user about applying critical damage


TimToxopeus:
Is this something we only want to inform the player who is applying the damage? Or do we want to call it out in chat for everyone to see? Might be a better experience if this is called out publicly, or at least following the chat settings, so the GM can hide critical damage if they so wish to.

At least all players will see "Player Ship took 120 damage, crossing # crit thresholds!" and can begin freaking out. Maybe we can even add a chat card button to roll for the damaged system, and automatically adjust it?

danimrath:
Consolidating changes into https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/pull/1415.

PR #651, opened 2022-06-02
Add system setting to automatically roll starship crit effects

LebombJames:
When enabled, when a starship crosses at least one critical threshold, the critical damage table will automatically be rolled (multiple times if multiple CTs were crossed). Damage must be applied from the chat card in order for this to work.

(Will be updated to include #650 when that's merged)


TimToxopeus:
Maybe instead of making this automatic, we can introduce a setting for it to be automatic or manual, and require the user to click a "Roll crit effect" button on the "Ship X took Y damage, crossing Z thresholds" card button? See #650 for this suggestion.

danimrath:
Consolidating changes into https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/pull/1415.

PR #652, opened 2022-06-03
Added critical status automation to starship weapon attack buttons

LebombJames:
Currently, the only way to get critical status automation on starship weapon attacks (e.g -2 on glitching, and -2 on malfunctioning power core) is to use the Shoot action. This change adds that automation to the Attack buttons on the weapon tab, as (at least I have found with my group) they use those over the Shoot action.


LebombJames:
There's definitely something to be done about the Shoot action vs attack buttons, but I realized there's a much, *much* easier solution for the meantime.

danimrath (commented):
This is already in the dev branch.

PR #656, opened 2022-06-06
Floating Numbers on HP/SP change

LebombJames:
Implementation of @mkahvi's floating damage numbers module into the system.
Configuration settings are included in their own submenu on the system settings page in order to keep things tidy. If there's a better way to lay that out in settings.js then let me know.
Same goes for the functions in actor.js, didn't know where else to put them.


mkahvi:
Posting here as well, for v10 support:

Replace
```js
t.hud.createScrollingText(label, options);
```
with
```js
canvas.interface.createScrollingText(t.center, label, options);
```

PR #657, opened 2022-06-06
Negative Level Actor Resource

LebombJames:
Made a Negative Level Actor Resource, where the value is the number of negative levels you have. Minor issue is that there are currently no modifiers to current HP/SP. If this isn't the approach we wanna take with -ve Levels that's fine, just wanted to put the idea out there.


TimToxopeus:
That's precisely what I had in mind for the actor resources when I built the system! :)

PR #663, opened 2022-06-11
Chat roll colors issue 457

### LINKED ISSUE #457: Chat Roll colors do not change depending on roll type
In the Chat Window all roll types always appear blue, regardless of the chosen roll type.
Therefore, it is not possible for a GM to determine if a Player's roll was a public roll, or Private/Blind GM roll.

Tested in Starfinder 0.14.3, FVTT 0.8.9, with no active modules.

SF_Roll_Colors

avagdumortesin:
Resolves #457

### DIFF
--- a/src/less/chat.less
+++ b/src/less/chat.less
@@ -11,12 +11,12 @@
   position: relative;
 }
 #chat-log .message.whisper {
-  background: #e8e8ef;
-  border: 5px solid #545469;
+  background-color: #7076bf66;
+  border: 2px solid #7096bf;
 }
 #chat-log .message.blind {
-  background: #ffebff;
-  border: 5px solid #6b476b;
+  background-color: #ffcafd66;
+  border: 2px solid #d87fd6;
 }
 #chat-form {
   textarea {
@@ -142,16 +142,6 @@
     z-index: 1;
     color: @sfrpgRed;
     font-weight: bold;
-  }
-  .message-metadata {
-    flex: 2;
-    text-align: right;
-    font-size: 12px;
-  }
-  .flavor-text,
-  .whisper-to {
-    flex: 0 0 100%;
-    font-size: 12px;
   }
 }
 
--- a/src/less/chat.less
+++ b/src/less/chat.less
@@ -18,16 +18,6 @@
   background: #ffebff;
   border: 5px solid #6b476b;
 }
-.chat-message .message-header .message-metadata {
-  flex: 2;
-  text-align: right;
-  font-size: 12px;
-}
-.chat-message .message-header .flavor-text,
-.chat-message .message-header .whisper-to {
-  flex: 0 0 100%;
-  font-size: 12px;
-}
 #chat-form {
   textarea {
     background: @sheetBackground;
@@ -153,7 +143,18 @@
     color: @sfrpgRed;
     font-weight: bold;
   }
+  .message-metadata {
+    flex: 2;
+    text-align: right;
+    font-size: 12px;
+  }
+  .flavor-text,
+  .whisper-to {
+    flex: 0 0 100%;
+    font-size: 12px;
+  }
 }
+
 
 .sfrpg {
   &.chat-card {
--- a/src/less/chat.less
+++ b/src/less/chat.less
@@ -10,7 +10,24 @@
   font-size: 13px;
   position: relative;
 }
-
+#chat-log .message.whisper {
+  background: #e8e8ef;
+  border: 5px solid #545469;
+}
+#chat-log .message.blind {
+  background: #ffebff;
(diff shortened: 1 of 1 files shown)


src/less/chat.less
```diff
@@ -10,7 +10,14 @@
   font-size: 13px;
   position: relative;
 }
-
+#chat-log .message.whisper {
+  background: #e8e8ef;
```
wildj79, line 14: This needs to be `background-color` instead of just `background`. We use a transparent `.webp` image on the background, so any color you set will show through the image. Using `background` is overriding that image, and we need to keep the image to maintain consistency with the rest of the design.

Also, the chosen color doesn't really work in this context. I realize you just copied the hex code being used in foundry core, but it doesn't stand out against the Starfinder color scheme very well. It's a little too subtle. We need something that's going to say "I'm obviously different form the other chat cards".

src/less/chat.less
```diff
@@ -10,7 +10,14 @@
...
+  border: 5px solid #545469;
```
wildj79, line 15: I'm not sold on the thicker border here. I realize it ups the visibility of the chat card quite a bit, but it adds inconsistency with the overall design of the system.

Same critique with the chosen color here as above. It just blends into the system color scheme a little to much. This needs to be a color that is gonna stand out against the systems color's (without being overly overt, either).

src/less/chat.less
```diff
@@ -10,7 +10,14 @@
...
+}
+#chat-log .message.blind {
+  background: #ffebff;
```
wildj79, line 18: Same as above, use `background-color` instead of just `background`. 

The color here is better, since it stands out and the chat card is obviously different from the norm. We might play with it though, and see if we can find a better color. Overall, this one is fine, though.

src/less/chat.less
```diff
@@ -10,7 +10,14 @@
...
+  border: 5px solid #6b476b;
```
wildj79, line 19: Still not liking the thicker border here. But, as above, the color is fine, if we can't find another color to take its place.

src/less/chat.less
```diff
@@ -136,8 +143,19 @@
     color: @sfrpgRed;
     font-weight: bold;
   }
+  .message-metadata {
```
wildj79, line 146: This is redundant. Foundry core `css` already defines these, so they aren't doing anything other than masking other styling that's already there.

src/less/chat.less
```diff
@@ -136,8 +143,19 @@
...
+    flex: 2;
+    text-align: right;
+    font-size: 12px;
+  }
+  .flavor-text,
+  .whisper-to {
```
wildj79, line 152: Same with this block. These are already defined in core `css` styling so aren't needed here.

wildj79 (changes requested):
Hey, thanks for the contribution! I've highlighted a few issues that I see with this PR. If you can correct these, we'll see about getting this in for the next release.

src/less/chat.less
avagdumortesin: I think https://colour-palette-generator.azurewebsites.net/?r=112&g=150&b=191 might work for the border here; it's a complementary color to the header background color (basically just a darker version).

src/less/chat.less
avagdumortesin: Looking at the blue for the header (https://colour-palette-generator.azurewebsites.net/?r=150&g=200&b=255), we can definitely find some complementary colors to that. For this color being a gray, there's not a good complementary gray, but https://colour-palette-generator.azurewebsites.net/?r=112&g=118&b=191 might work, and should look good with the border below.

src/less/chat.less
avagdumortesin: https://colour-palette-generator.azurewebsites.net/?r=255&g=202&b=253 is more complementary to the header color.

src/less/chat.less
avagdumortesin: https://colour-palette-generator.azurewebsites.net/?r=216&g=127&b=214

src/less/chat.less
avagdumortesin: I think these were leftover from when I was trying to figure out why the `.whisper` and `.blind` text wasn't appearing on the GM's view.

src/less/chat.less
avagdumortesin: I think these were leftover from when I was trying to figure out why the `.whisper` and `.blind` text wasn't appearing on the GM's view.

avagdumortesin:
Screen Shot 2022-06-12 at 12 45 38 PM
Here's what the colors I suggested above look like with your suggested changes.

wildj79 (approved):
Ok, these look pretty good. I think we're good for merge now.

PR #712, opened 2022-08-07
Prettify roll formulas on chat cards

LebombJames:
image

First card is with the change, second is without.

`roll.formula` produces a cleaned up formula already, whereas `roll._formula` doesn't, so changing chat cards to use the former is a very easy fix.


TimToxopeus:
Nice clean-up! :)

PR #725, opened 2022-09-03
V10: Drag and Drop and other misc fixes

LebombJames:
(remade PR because git decided we needed to bring this branch up to date with dev)
-Moved HTML template preload to init because sheets refuse to open otherwise.
-Explicitly add `isOpen` to containers, as it didn't seem to be adding before, fixing containers not opening or closing
-Refactor drag-drop to actor sheets to use Item.fromDropData, in the process removed some redundant logic, like if the item, came from a compendium or the sidebar. I've managed to get containers 99% functional, there are some minor bugs: some items refuse to stack, and the console spits out all sorts of messages about unknown items, even though the items are being moved fine.

### DIFF
--- a/src/module/actor/actor-inventory-utils.js
+++ b/src/module/actor/actor-inventory-utils.js
@@ -215,7 +215,7 @@ export async function moveItemBetweenActorsAsync(sourceActor, itemToMove, target
             }
 
             const duplicatedData = duplicate(itemToCreate.item);
-            if (duplicatedData.data.equipped) {
+            if (itemToCreate.item.data.equipped) { //duplicate() doesn't duplicate .data getter
                 duplicatedData.data.equipped = false;
             }
             
@@ -247,7 +247,9 @@ export async function moveItemBetweenActorsAsync(sourceActor, itemToMove, target
         }
 
         /** Ensure the original to-move item has the quantity correct. */
-        itemData[0].data.quantity = quantity;
+        if (itemData[0].system?.quantity) { //Check for V10
+            itemData[0].system.quantity = quantity; //.data isn't on duplicated objects, so no easy way to do this on V9
+        }
 
         if (itemData.length != items.length) {
             console.log(['Mismatch in item count', itemData, items]);
--- a/src/module/actor/actor-inventory-utils.js
+++ b/src/module/actor/actor-inventory-utils.js
@@ -247,8 +247,10 @@ export async function moveItemBetweenActorsAsync(sourceActor, itemToMove, target
         }
 
         /** Ensure the original to-move item has the quantity correct. */
-        if (itemData[0].system?.quantity) { //Check for V10
-            itemData[0].system.quantity = quantity; //.data isn't on duplicated objects, so no easy way to do this on V9
+        if (isNewerVersion(game.version, '10.0')) {
+            itemData[0].system.quantity = quantity;
+        } else {
+            itemData[0].data.quantity
         }
 
         if (itemData.length != items.length) {
--- a/src/module/actor/actor-inventory-utils.js
+++ b/src/module/actor/actor-inventory-utils.js
@@ -215,9 +215,7 @@ export async function moveItemBetweenActorsAsync(sourceActor, itemToMove, target
             }
 
             const duplicatedData = duplicate(itemToCreate.item);
-            if (itemToCreate.item.data.equipped) { //duplicate() doesn't duplicate .data getter
-                duplicatedData.data.equipped = false;
-            }
+            duplicatedData.system.equipped = false;
             
             items.push({item: duplicatedData, children: contents, parent: itemToCreate.parent});
         }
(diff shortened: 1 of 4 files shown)


src/module/actor/actor-inventory-utils.js
```diff
@@ -247,7 +247,9 @@ export async function moveItemBetweenActorsAsync(sourceActor, itemToMove, target
         }
 
         /** Ensure the original to-move item has the quantity correct. */
-        itemData[0].data.quantity = quantity;
+        if (itemData[0].system?.quantity) { //Check for V10
```
wildj79, line 250: This is problematic at the moment. The goal for this update is to not break v9 comparability. We'll need to make use of a version check here to determine how this block of code needs to be handled. I don't like doing it, but it's gonna be the only way to handle v9 and v10 compatibility for this one.

src/module/actor/sheet/base.js
```diff
@@ -821,6 +821,7 @@ export class ActorSheetSFRPG extends ActorSheet {
         await actorHelper.updateItem(item.id, update);
 
         const itemData = duplicate(item.data);
+        itemData.data = item.data.data; //duplicate doesn't copy .data getter, so we have to graft it on manually.
```
wildj79, line 824: This is another instance where we'll need a version check before handling this block. The problem is that you're making your code changes in a v10 world, where this comment is true, but it won't be true on v9. So, we have to check what version of foundry is running, and then populate the data items accordingly.

src/module/actor/sheet/base.js
```diff
@@ -987,6 +988,24 @@ export class ActorSheetSFRPG extends ActorSheet {
             return;
         }
 
+        const itemData = await Item.fromDropData(parsedDragData);
```
wildj79, line 991: Need to check if this method exists in v9 or not. If not, then we'll need to do more version checking before processing.

src/module/actor/sheet/base.js
```diff
@@ -1010,58 +1029,39 @@ export class ActorSheetSFRPG extends ActorSheet {
                 ui.notifications.warn(game.i18n.format("SFRPG.ActorSheet.Inventory.Interface.ItemCollectionPickupNoGMError"));
             }
             return;
-        } else if (parsedDragData.pack) {
```
wildj79, line 1013: We can't completely rip these lines out yet. They might have to stay in until we've completely updated the system for v10 compatibility.

wildj79 (changes requested):
So, you'll notice that the theme of this review is that we need to do some version checking before updating certain blocks of code. As it stands right now, these changes will completely break v9 compatibility. So, we have two options, either we hold off on merging this in until we're ready for the v10 upgrade, or we add in the version checks and handle the code blocks accordingly.

src/module/actor/sheet/base.js
LebombJames: Can confirm it does

src/module/actor/sheet/base.js
LebombJames: Added version checks to these to ensure they function on v9.

LebombJames:
Everything should be taken care of now. With this and #726 V10 should be usable. I must say I'm concerned about the future work we're creating for ourselves here in stripping this all out later when we fully target V10.

src/module/actor/sheet/base.js
wildj79: Unfortunately, this method doesn't work when the item type is of type `ItemCollection`. This is a custom type that we use with the loot sheets. So, you can drag an item to the canvas, but can't drag that item from the loot sheet onto an actor.

src/module/actor/sheet/base.js
```diff
@@ -1010,58 +1031,41 @@ export class ActorSheetSFRPG extends ActorSheet {
...
             const itemInTargetActor = await moveItemBetweenActorsAsync(targetActor, addedItem, targetActor, targetContainer);
             if (itemInTargetActor === addedItem) {
                 await this._onSortItem(event, itemInTargetActor.data);
                 return itemInTargetActor;
             }
 
             return itemInTargetActor;
-        } else if (parsedDragData.data) {
-            const sourceActor = new ActorItemHelper(parsedDragData.actorId, parsedDragData.tokenId, parsedDragData.sceneId);
+        } else if (parsedDragData.uuid.includes("Actor") || parsedDragData.data) {
```
wildj79, line 1054: `uuid` isn't a property in the drag data on v9, so this throws an error when you drag items onto an actor. a `?` after `uuid` should fix this.

wildj79 (changes requested):

wildj79:
> h

Normally, I'd agree with you. I'm generally against version checks because it muddies the code base. But, in this instance, for this one release, I don't really see a way around it. Once v10 has been out for a few weeks, and the system is running OKish on it, then we can focus on v10 specific programming.

wildj79:
After considering things for a bit, I think we just might need to bite the bullet and not worry about v9 compatibility. I can do a v9 release and then do a separate one that puts in the bare minimum fixes needed for v10 to work.

LebombJames:
@wildj79 So what's the plan wrt to the V9 update? Should I adjust this merge to fully target V10, or to be compatible on V9 too?

wildj79 (approved):

wildj79:
Ok, I'm going to go ahead and merge these in now. There's no going back now, we're locked into v10 from this point forward ;) .

LebombJames:
Still to do: the `ItemCollection` drag data stuff, and some compendium to loot actor bugs

PR #726, opened 2022-09-04
V10: Canvas Drop handler fixes

LebombJames:
Loot actor sheets still behave oddly. This fix just restores dropping items to the canvas using a Hook instead of monkey patching core (which mainly consists of taking the existing functions, and rejigging them to work on the hook and on V10). Isolated it to its own function too.


wildj79:
Merging this in now. I'm gonna run some tests locally. If everything checks out, I'll issue a new release.

LebombJames:
This also needs its V9 version checking removed

PR #733, opened 2022-09-18
V10: Fix text fields always showing the editor page

LebombJames:
Currently, text fields (such as the biography on an actor) always show the editor page, even after being saved. This merge fixes that by properly migrating to the new `editor` handlebar helper (and stripping out whatever `editorPlus` was, since the new `editor` handlebar helper does all `editorPlus` did). This is still using TinyMCE; we could migrate to ProseMirror if we want.

### DIFF
--- a/src/module/actor/sheet/character.js
+++ b/src/module/actor/sheet/character.js
@@ -27,7 +27,7 @@ export class ActorSheetSFRPGCharacter extends ActorSheetSFRPG {
         return path + "character-sheet.html";
     }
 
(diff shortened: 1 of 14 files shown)


wildj79:
I got a little concerned when I saw that all of the `getData()` methods were being made `async`, but it appears that this is supported in foundry at the moment, so I'm OK with it. I'm going to go ahead and merge this for now with the understanding that I might revert it if things start to go wonky because of the `async` functions.

PR #741, opened 2022-09-21
V10: Fix bulk and wealth

LebombJames:
Currently total wealth shows NaN, and Bulk/the encumbrance bar is missing. This merge restores correct wealth calculation, and display of bulk and the encumbrance bar.


wildj79:
Looks good, merging.

PR #742, opened 2022-09-21
V10: Fix cook replacing journal pages with mystery man

### LINKED ISSUE #740: [Contribution bug] Cook outputs bugged journal entries.
Cook program outputs bugged journal entries from V10 exported journals. Cooked V10 journal items instead only displayed a Mystery_Man image in the entry, and nothing else.
This seems to be an issue with the Cook process adding lines of code that do not exist in V10 exported journal entries.
This only effects entries exported from V10, previous Journal items remain fine in the cook, and are read fine in Dev world.
(Have not yet tested other exported item types)

LebombJames:
Closes #740 

Currently, `cook` replaces a journal's content with a single image page with the mystery man icon. This is because cook sets `img` for any document without `img` (which journals in V10 by definition don't have) to mystery man. This merge changes that so that if the current document is a journal, it is skipped.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -350,7 +350,7 @@ async function cookWithOptions(options = { formattingCheck: true }) {
 
                 // Fix missing images
                 if (!jsonInput.img && !jsonInput.pages) {
-					// Skip if a journal
+                    // Skip if a journal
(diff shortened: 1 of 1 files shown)


wildj79:
Looks good, merging.

PR #751, opened 2022-10-05
Rules reformat

Iankid:
Updated Rules compendium to use V10 Journals.
Moved Skills from Skills compendium to Rules
Updated links to Rules and Skills through system.


Iankid:
Cleaned up

PR #753, opened 2022-10-07
Partly fix character sheets not opening after changes to base.js

LebombJames:
Currently, opening a character sheet throws an error that `data.items.sort` is not a function, because `data.items` is not an array. This change makes `data.items` an array. (character sheets still refuse to open, because we're trying to assign to getters with `hasAttack` and `hasDamage`. Not sure how it worked before, but it throws now 🤷‍♂️)

### DIFF
--- a/src/module/actor/sheet/base.js
+++ b/src/module/actor/sheet/base.js
@@ -74,7 +74,7 @@ export class ActorSheetSFRPG extends ActorSheet {
             config: CONFIG.SFRPG
         };
 
-        data.items = [...this.actor.items];
+        data.items = [...this.actor.items.values()];
         data.items.sort((a, b) => (a.sort || 0) - (b.sort || 0));
         data.labels = this.actor.labels || {};
         data.filters = this._filters;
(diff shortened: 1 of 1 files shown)


wildj79:
What happened is that what used to be in an outer `data` object has all been merged into the main object. So instead of `actor.data.someThing` it's `actor.someThing`. We just have to adjust a bit. I'm still mulling over the best way to handle the `hasAttack` and `hasDamage` issues (I was aware of them).

src/module/actor/sheet/base.js
```diff
@@ -74,7 +74,7 @@ export class ActorSheetSFRPG extends ActorSheet {
             config: CONFIG.SFRPG
         };
 
-        data.items = this.actor.items;
+        data.items = [...this.actor.items];
```
wildj79, line 77: Unfortunately this won't work. `actor.items` has been turned into a `Map`, so this is going to create an array of array's instead of an array of `Item`s. What you need to do is call the `Map.prototype.values()` method, so change it to:
```js
data.items = [...this.actor.items.values()];
```

wildj79 (changes requested):

src/module/actor/sheet/base.js
LebombJames: Oops, assumed `Map`s would spread as they were

wildj79 (approved):
Looks good.

PR #754, opened 2022-10-08
Whisper chat cards for hidden combatants

danimrath:
When a token in the combat tracker is hidden and it's turn comes up in the initiative this will stop it from announcing it to everyone automatically by whispering to the GM.

### DIFF
--- a/src/module/combat/combat.js
+++ b/src/module/combat/combat.js
@@ -481,7 +481,7 @@ export class CombatSFRPG extends Combat {
         const chatData = {
             type: CONST.CHAT_MESSAGE_TYPES.OTHER,
             speaker: ChatMessage.getSpeaker({ actor: eventData.newCombatant, token: eventData.newCombatant?.token, alias: speakerName }),
-            whisper: eventData.newCombatant.hidden ? [game.user.data._id] : [],            
+            whisper: eventData.newCombatant.hidden ? [ChatMessage.getWhisperRecipients("GM")] : [],            
             content: html
         };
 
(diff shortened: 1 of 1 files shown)


src/module/combat/combat.js
```diff
@@ -481,6 +481,7 @@ export class CombatSFRPG extends Combat {
         const chatData = {
             type: CONST.CHAT_MESSAGE_TYPES.OTHER,
             speaker: ChatMessage.getSpeaker({ actor: eventData.newCombatant, token: eventData.newCombatant?.token, alias: speakerName }),
+            whisper: eventData.newCombatant.hidden ? [game.user.data._id] : [],
```
wildj79, line 484: This should use `ChatMessage.getWhisperRecipients("GM")` instead of using `[game.user.data._id]`.

wildj79 (changes requested):

src/module/combat/combat.js
danimrath: Awesome thanks! Done!

src/module/combat/combat.js
```diff
@@ -481,7 +481,7 @@ export class CombatSFRPG extends Combat {
         const chatData = {
             type: CONST.CHAT_MESSAGE_TYPES.OTHER,
             speaker: ChatMessage.getSpeaker({ actor: eventData.newCombatant, token: eventData.newCombatant?.token, alias: speakerName }),
-            whisper: eventData.newCombatant.hidden ? [game.user.data._id] : [],            
+            whisper: eventData.newCombatant.hidden ? [ChatMessage.getWhisperRecipients("GM")] : [],
```
wildj79, line 484: The method call doesn’t need to be wrapped in brackets, since it already returns an array of user ids.

wildj79 (changes requested):

wildj79 (approved):
There we go, that looks good.

src/module/combat/combat.js
danimrath: My apologies!

PR #752, opened 2022-10-07
Restore usage buttons

LebombJames:
Closes #737

PR #760, opened 2022-10-19
Restore damage buttons and charged text

LebombJames:
Reverts part of #752 (data -> system changes fixed it anyway) and fixes broken localisation string


wildj79:
Good catch. Merging.

PR #761, opened 2022-10-19
Restore chat card descriptions

LebombJames:
Currently, chat cards of items (assumably spells too) have no descriptions due to the recent migration.

(reverting it to `data` could've worked, but going through `item.system` does too; they're both on the chat card data object)

### DIFF
--- a/src/templates/chat/item-card.html
+++ b/src/templates/chat/item-card.html
@@ -5,13 +5,13 @@
     </header>
 
     <div class="card-content">
-        {{#if system.description.short}}
-            {{{system.description.short}}}
+        {{#if item.system.description.short}}
+            {{{item.system.description.short}}}
         {{else}}
-            {{{system.description.value}}}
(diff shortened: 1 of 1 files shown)


wildj79:
I think for this one, the better option would be to change the underlying data structure in `ItemSFRPG#roll()` rather than touching the template. Basically, change it like this:

```js
const templateData = {
    actor: this.actor,
    tokenId: token ? `${token.parent.id}.${token.id}` : null,
    item: this,
    system: this.getChatData(htmlOptions), // <-- changed from data to system
    labels: this.labels,
    hasAttack: this.hasAttack,
    hasDamage: this.hasDamage,
    hasSave: this.hasSave,
    hasOtherFormula: this.hasOtherFormula
};
```

It's a little more consistent with the rest of the data model changes and is less confusing overall. I'm actually working on this right now, so I'm just gonna close this PR.

PR #768, opened 2022-10-25
Fix missing add spell button if all spells are innate/always available

### LINKED ISSUE #766: missing Add button if all spells are innate
In a sheet's spellbook, if all spells are innate casting, the add spell button disappears.

LebombJames:
Closes #766 

This has actually been a bug for a while 😉


wildj79:
Simple fix. Merging.

PR #769, opened 2022-10-25
Improve readability of private and blind chat cards 

LebombJames:
Public (for reference)
image

Private
image

Blind
image


wildj79:
Yep, that looks a lot better. Merging.

PR #771, opened 2022-10-26
Fix Starship weapon attack bonus box

### LINKED ISSUE #728: starship attack modifier not working.
Hello, it appears that the starship attack modifier does not work efficiently :
issue

Also : Could you add some Starship specific Modifier effects, starship specific. E.G. Starship AC, since all of the current modifier options are for PC/NPC sheets at the moment.

LebombJames:
Currently the attack bonus box on starships doesn't work
Closes #728


wildj79:
This one is a fit for "lowest hanging fruit", so I'm gonna go ahead and merge it for the next release.

PR #772, opened 2022-10-26
Fix broken headings on chat cards

LebombJames:
Before
image

After
image


wildj79:
Another one fit for "lowest hanging fruit", so mergin'. (Though, honestly, I thought I had already fixed this one at one point 🤷 ).

PR #776, opened 2022-10-29
Magic hybrid major update & rules reformat v10

### LINKED ISSUE #719: Missing Magic Items marked as Worn
This is a report to myself, and from anyone who can reply to this.
There are magic items and hybrid, both already in the compendium and not, that are not marked as "Worn" in SRD sites. But are in their books. 
I will be able to access the core line of rulebooks to learn about "Worn" magic and hybrid items from those. But magic and hybrid items from APs, Single adventures, and SFS, I will need replies from people who own them, to list off "Worn" magic and hybrid items, as displayed in your books.

Iankid:
Update to many Hybrid and Magic items. 
Includes filling "Worn" checkbox for relevant items.
Reassigning items to Hybrid or Magic from other types.

Added several missing Rules entries.
Reformatted rules for use in V10.
Removed Skills Compendium, and added Skills to SFRPG Rules Compendium
Updated links in many items leading to SFRPG Rules and Skills.

Fixes #719


LebombJames:
Some of the aliens modified are erasing the migration to `system`, and are still using `data`

wildj79:
@LebombJames I added a script in the `scripts` folder that we can run after merging this and before cooking to update them.

LebombJames:
I think it'd be best if @Iankid ran the migration on the branch, so we can merge it in cleanly, and see we can see if any other changes have been affected (ideally a rebase would fix it, but that could get messy with the timeline of things and the number of changes)

Iankid:
I would need to know which script, and anything special I need to do in regards to it. But if @wildj79 is ok with it, sure.

wildj79:
The script is located at `scripts/migrations/convert_all_files_to_v10_datamodel.mjs`. To run it, open a command prompt, `cd` into the root directory of the project, and then use this to call:

```
node scripts/migrations/convert_all_files_to_v10_datamodel.mjs
```

Should be that simple.

LebombJames:
Looks good now, last thing: delete the section about skills.db starting at line 137 in system.json

PR #787, opened 2022-11-02
Fix opening ASI sheets

### LINKED ISSUE #786: Ability Score Increase
Hello, 
the other day playing they went up to level 10, and when we hit ADD for Ability Score Increase: Level 10, when i try edit, don't work, don't open a window for chose. 

Thanks.

LebombJames:
Closes #786


Carantin:
Thanks

PR #819, opened 2022-11-22
Creature Companions

SCMNomad:
Created sheets for Creature Companions, excluding IS companions


LebombJames (changes requested):
I know @Iankid helped you with this so he can help you fix the issues; should just be a matter of bulk find+replace

PR #835, opened 2022-12-03
Skill Expertise Update

### LINKED ISSUE #834: Include modifiers into Skill Expertise (ex) Envoy class feature
During my first time using this system and after importing a character from hephaistos I was confused with how I should be handling the Envoy class feature `Skill Expertise (ex)` (https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/blob/development/src/items/class-features/skill_expertise_(ex).json).

```text
At 1st level and every 4 levels thereafter, you can use expertise with one additional class skill. You must have at least 1 rank in a skill to select it, and it must come from the following list: Bluff (Cha), Computers (Int), Culture (Int), Diplomacy (Cha), Disguise (Cha), Engineering (Int), Intimidate (Cha), and Medicine (Int).
```

I was explained on Discord the modifiers reflecting those expertise should be added to the class feature modifiers. This is a cumbersome process with fairly advanced formulas which I wouldn't expect my players to do on their own.

Those modifiers could be included into the class feature by default, in disabled state, and a player would simply enable them as his character level up.

**Additional context**
As an example, the `Expertise (ex)` class feature comes with a [built-in modifier](https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/blob/development/src/items/class-features/expertise_(ex).json#L45). This could be replicated for the skill expertise class feature.

Iankid:
-Updated skill expertise related items with modifiers, and making them active feats for forgoing skill expertise.
-Added missing Expert Advice (Ex) item.
Fixes #834


src/items/class-features/additional_skill_expertise_(ex).json
```diff
@@ -35,18 +40,180 @@
...
+      {
+        "_id": "d07b67b1-2c1b-4cc2-95fd-732f4e18085d",
+        "name": "Additional Skill Expertise (Ex)",
+        "type": "insight",
+        "condition": "",
+        "effectType": "skill",
+        "enabled": false,
+        "max": 6,
+        "modifier": "1d(lookupRange(@classes.envoy.levels,6,13,8))+lookupRange(@classes.envoy.levels,0,5,1,9,2,17,3,20,4)",
+        "modifierType": "constant",
```
LebombJames, line 105: These need to be situational in order for the dice to work correctly. Additionally, if they're situational you don't need to make it an active feat, you can just turn it off when needed

LebombJames (changes requested):

Iankid:
This PR awaiting "Null" option in skill modifiers for better user interface.

src/items/class-features/hunter’s_expertise_(ex)_(1st).json
```diff
@@ -30,22 +32,95 @@
...
+        "_id": "a34f83bd-b6f4-4f5a-8255-af61fb8f6674",
+        "name": "Hunter’s Expertise (Ex) - Secondary fighting style",
+        "type": "untyped",
+        "condition": "",
+        "effectType": "skill-points",
+        "enabled": false,
+        "max": -8,
+        "modifier": "(@classes.soldier.levels - 8)",
+        "modifierType": "constant",
+        "notes": "You gain a free skill rank at each soldier level, from 9th onward, you can use only for Perception or Survival (this does not allow you to exceed the maximum number of skill ranks in a single skill).",
```
LebombJames, line 118:
```suggestion
        "notes": "If you take the hunter fighting style as your secondary fighting style, from 9th onward you gain a free skill rank at each soldier level you can use only for Perception or Survival (this does not allow you to exceed the maximum number of skill ranks in a single skill).",

PR #836, opened 2022-12-03
IS creature companions

SCMNomad:


src/items/creature-companions/linnormling.json
```diff
@@ -0,0 +1,1523 @@
...
+        "uses": {
+          "max": 0,
+          "per": "",
+          "value": 0
+        }
+      }
+    }
+  ],
+  "prototypeToken": {
+    "name": "New Actor (54)",
```
LebombJames, line 912: Can you run cook on these so their prototype tokens are correct for the system settings? It should set the name and default settings correctly. Still don't commit the db files. If that doesn't do anything you'll need to rebase from development to get those changes to cook.

LebombJames (changes requested):

LebombJames:
Nice one 👍

PR #845, opened 2022-12-11
Ironmonk88 chat message changes

Dosendusche:
- resolved merge conflicts.

I think it would be good if someone could take a look if that would still work like this especially ironmonk88. The rolling in starfinder alone still works fine.


LebombJames:
How come this PR doesn't have the if statement around the chat message creation in dice.js?
image
image

LebombJames:
Wait nvm checking the full file I can see what's going on

src/module/dice.js
```diff
@@ -262,9 +262,9 @@ export class DiceSFRPG {
             const itemContext = rollContext.allContexts['item'];
             const htmlData = [{ name: "rollNotes", value: itemContext?.system?.rollNotes }];
 
-            let useCustomCard = game.settings.get("sfrpg", "useCustomChatCards");
+            let useCustomCard = game.settings.get("sfrpg", "useCustomChatCards") && chatMessage;
```
LebombJames, line 265: This check is needless given lines 267, 288, 757 and 783 check for `chatMessage` anyway

LebombJames (changes requested):

src/module/dice.js
Dosendusche: done. thanks for the hint.

src/module/dice.js
```diff
@@ -752,10 +752,10 @@ export class DiceSFRPG {
             htmlData.push({ name: "damage-parts", value: JSON.stringify(tempParts) });
             htmlData.push({ name: "rollNotes", value: itemContext?.data?.damageNotes });
 
-            let useCustomCard = game.settings.get("sfrpg", "useCustomChatCards");
+            let useCustomCard = game.settings.get("sfrpg", "useCustomChatCards") && chatMessage;
```
LebombJames, line 755: Didn't spot this one too, my bad

LebombJames (changes requested):

src/module/dice.js
Dosendusche: yep, same. I could have done it before.

Issue #848, opened 2022-12-11
Temp HP is always deleted when applying damage, and causes stamina to behave oddly


LebombJames:
**Describe the bug**
When applying damage to actors with Temp HP, the Temp HP is always reduced to 0 no matter the damage, additionally, stamina seems to be changed by (the number of Temp HP the actor had - the incoming damage), which can cause it to go up or down.

**To Reproduce**
Apply damage to an actor with Temporary HP

**Expected behavior**
The Temp HP should be subtracted by the amount of incoming damage

**Screenshots**

**Technical details (please complete the following information):**
 - OS: Windows 10
 - Browser: Chome
 - Foundry Version: 10.291
 - SFRPG Version: 0.22.0

**Additional context**
Also occurs on 0.21.0

Dosendusche:
I also had this problem in one of my sessions. Kinda personal interest.
I will take a look at that.

PR #850, opened 2022-12-12
temp hp damage calculation

Dosendusche:
- fixed temp hp

Fixes https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/issues/848

### DIFF
--- a/src/module/actor/mixins/actor-damage.js
+++ b/src/module/actor/mixins/actor-damage.js
@@ -320,20 +320,21 @@ export const ActorDamageMixin = (superclass) => class extends superclass {
 
         if (!damage.isHealing) {
             /** Update temp hitpoints */
-            let newTempHP = Math.clamped(originalTempHP - remainingUndealtDamage, 0, actorData.attributes.hp.tempmax);
+            let newTempHP = Math.clamped(originalTempHP - remainingUndealtDamage, 0,
+                actorData.attributes.hp.tempmax ? actorData.attributes.hp.tempmax : actorData.attributes.hp.temp);
             remainingUndealtDamage = remainingUndealtDamage - (originalTempHP - newTempHP);
 
             if (newTempHP <= 0) {
                 newTempHP = null;
                 actorUpdate['system.attributes.hp.tempmax'] = null;
             }
-            
+
             actorUpdate["system.attributes.hp.temp"] = newTempHP;
 
             /** Update stamina points */
             const newSP = Math.clamped(originalSP - remainingUndealtDamage, 0, actorData.attributes.sp.max);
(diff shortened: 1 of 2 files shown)


Dosendusche:
Converted to draft as I might delete the `+Max` UI field and it's functionality completely from the damage calculation.

src/module/actor/mixins/actor-damage.js
```diff
@@ -320,20 +320,21 @@ export const ActorDamageMixin = (superclass) => class extends superclass {
 
         if (!damage.isHealing) {
             /** Update temp hitpoints */
-            let newTempHP = Math.clamped(originalTempHP - remainingUndealtDamage, 0, actorData.attributes.hp.tempmax);
+            let newTempHP = Math.clamped(originalTempHP - remainingUndealtDamage, 0,
+                actorData.attributes.hp.tempmax ? actorData.attributes.hp.tempmax : actorData.attributes.hp.temp);
```
LebombJames, line 324:
```suggestion
let newTempHP = Math.clamped(originalTempHP - remainingUndealtDamage, 0, actorData.attributes.hp.tempmax || actorData.attributes.hp.temp);
```
This can be done a bit more concisely

LebombJames (commented):
Just something I spotted. Not sure why this PR fixes the bug but sure 😄

src/module/actor/mixins/actor-damage.js
Dosendusche: You are totally correct but I discovered it would be better to just remove the `+Max` field from the UI and not calculate with it anymore. Because it's not used in the starfinder system anyway. But since there are around 1400 lines in the code (mostly alien archive) having that property completely removing that thing from the code should be a different PR or maybe never done at all.

Dosendusche: > You are totally correct but I discovered it would be better to just remove the `+Max` field from the UI and not calculate with it anymore. Because it's not used in the starfinder system anyway. But since there are around 1400 lines in the code (mostly alien archive) having that property completely removing that thing from the code should be a different PR or maybe never done at all.

nvm found something that heals to a specific max amount of temp hp: https://aonsrd.com/ArmorUpgrades.aspx?ItemName=Brown&Family=Force%20Field
There might be more but this one is evidence enough.
Since there is currently nothing automated that heals tempHP to a maximum it's just not needed.
But if that ever gets automated the `hp.maxtemp` should be set inside the item/feat/object whatever that is providing heal to a specific max temp hp. I will update this PR accordingly to my understanding of what is needed. If anyone disagrees with me here we can happily discuss or maybe just leave it as is.

LebombJames:
We'll need to think about removing `tempmax` from actors via a world migration/a script on the alien-archives

PR #853, opened 2022-12-13
Damage dialog text color fix

Iankid:
Removed color override from Damage dialog window. Allowing compatibility with UI mods.

### DIFF
--- a/src/less/roll.less
+++ b/src/less/roll.less
@@ -9,7 +9,6 @@
 
   div.damage-type-selector-list {
     .damage-type-list-title {
-      color: @colorOlive;
       font-weight: bold;
       font-size: 1em;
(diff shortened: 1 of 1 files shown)


src/less/roll.less
```diff
@@ -9,15 +9,13 @@
...
-      color: @colorOlive;
       font-weight: bold;
       font-size: 1em;
       border-bottom: 1px solid @borderDark;
       margin-top: 5px;
     }
 
     .damage-type {
-      color: @colorOlive;
     }
```
LebombJames, line 19:
```suggestion
```
Don't leave empty rules

LebombJames (changes requested):

PR #866, opened 2022-12-17
add action buttons in item collections

Dosendusche:
- enabled item action buttons inside item collections

This was requested by @Iankid to make solar crystals work inside a Solar Weapon.
Just a more flexible approach to allow all items to have their buttons.

### DIFF
--- a/src/less/variables.less
+++ b/src/less/variables.less
@@ -98,3 +98,7 @@
     .flex3 { flex: 3; }
     .flex4 { flex: 4; }
 }
+
+.flexnowrap {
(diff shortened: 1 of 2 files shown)


LebombJames:
So currently with both attack and damage buttons, this causes small items to be overfull.
image


(Although its fine with only one button)
image

Playing around a bit, changing the flex-direction of the buttons to `row` and setting their margins to 0 creates a nice alternative
image

Dosendusche:
thx for keeping track of the styling stuff. I will commit a change.

PR #871, opened 2022-12-18
compendium browser buttons for actorsheet

Dosendusche:
- added pre filtered compendium browser buttons to actorsheet (inventory and spellbook)
- added handlebars helper for lazy people like me to debug templates easier
- added some localization text

Some more frontend related people might want to check the button styling. I have now idea how to make that look good

### DIFF
--- a/src/module/actor/sheet/base.js
+++ b/src/module/actor/sheet/base.js
@@ -11,6 +11,7 @@ import { InputDialog } from "../../apps/input-dialog.js";
 import { SFRPG } from "../../config.js";
 
 import { ItemSFRPG } from "../../item/item.js";
+import { getEquipmentBrowser } from "../../packs/equipment-browser.js";
 /**
  * Extend the basic ActorSheet class to do all the SFRPG things!
  * This sheet is an Abstract layer which is not used.
@@ -218,6 +219,9 @@ export class ActorSheetSFRPG extends ActorSheet {
 
         // Create New Item
         html.find('.item-create').click(ev => this._onItemCreate(ev));
+
+        // Get New Item from Browser
+        html.find('.item-browser').click(event => this._onOpenBrowser(event));
 
         // Update Inventory Item
         html.find('.item-edit').click(ev => {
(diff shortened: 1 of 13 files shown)


LebombJames:
Unfortunately the only section that works for me is Weapons, all else open without filters. I don't know what else you changed, or if this is a bug on my end (somehow) 🤷‍♂️

src/sfrpg.js
```diff
@@ -665,6 +665,10 @@ function setupHandlebars() {
         const formattedValue = moneyFormatter.format(value);
         return formattedValue;
     });
+
+    Handlebars.registerHelper('json', function(context) {
```
LebombJames, line 669: Also how does this differ from the built-in `{{log}}` function?

src/templates/actors/parts/actor-spellbook.hbs
```diff
@@ -51,6 +51,9 @@
             <a class="item-control item-create" title="{{localize "SFRPG.SpellBook.CreateSpell"}}" {{#each section.dataset as |v k|}}data-{{k}}="{{v}}"{{/each}}>
                 <i class="fas fa-plus"></i> {{localize "SFRPG.SpellBook.Add"}}
             </a>
+            <a class="item-control item-browser" title="{{localize "SFRPG.SpellBook.CreateFromCompendium"}}" {{#each section.dataset as |v k|}}data-{{k}}="{{v}}"{{/each}} data-classes="{{json section.classes}}">
```
LebombJames, line 54: Would also still like classes to just be their name, not a whole JSON

Iankid:
UI issue of note, is that the original add button is being shove awkwardly to the left. Also, a 
means to identify to two would be beneficial. Perhaps renaming the original text to "Add custom"?
addbutton

src/module/actor/sheet/base.js
```diff
@@ -556,6 +560,80 @@ export class ActorSheetSFRPG extends ActorSheet {
...
+        const filterType = event.currentTarget.dataset.type;
+        const classesToFilters = {
+            'mystic': 'myst',
+            'precog': 'precog',
+            'technomancer': 'tech',
+            'witchwarper': 'wysh'
+        };
+        let browser;
+        let activeFilters = {};
+        switch (filterType) {
```
LebombJames, line 574: Changing the switch case to an else if seems to fix equipment, but not spells. Again, no idea why this broke unless you changed something

src/sfrpg.js
Dosendusche: > Also how does this differ from the built-in `{{log}}` function?

funny didn't even know this `log` thing exists. I would say the main difference is this one prints the object directly onto the place it's put on as a string. Which made it possible for me to print the json from `section.classes` directly into the data attribute. 
But as this will now be changed this will not be needed anymore. Please decide if this should stay or not.

src/module/actor/sheet/base.js
Dosendusche: Problem solved. Switch case was to be changed and `browser.refreshFilters` needed to be set to true.

src/templates/actors/parts/actor-spellbook.hbs
Dosendusche: done.

Dosendusche:
> UI issue of note, is that the original add button is being shoved awkwardly to the left. Also, a means to identify between the two would be beneficial. Perhaps renaming the original text to "Add custom"? addbutton


Strange as this is what it looks for me:
Unbenannt
What is different between your and my installment? maybe we need to check this together @Iankid

LebombJames:
Looked the same for me. Did you commit your CSS?

Dosendusche:
yes, only change I did was the one in `actor.less` file which "fixed" it for me.
Still might need someone to do the css

LebombJames:
Ah, it's quick insert's fault

LebombJames:
image

Even looks good with Quick Insert!
image

PR #880, opened 2022-12-21
Modifier system fixes

Dosendusche:
- [added formula calculation on drop and toggle](https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/commit/ca51cb164bb7a137138a84cfc207bbe62a9bb193)
- [small chatbox bugfix, added function docs](https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/commit/552ff7a78f90462f2b354259fd4d3c479a32b02b)
- [fixed formula calculation for drop and sheet](https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/commit/e6534933e0e4f5fd34c402bfe8e218227054c1ad)
- [added modifier stacking into dice rolling](https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/pull/880/commits/e65433c14f99b4850916d88eb78801245c4a98e5)
- [moved: roll stacking calc; added: createRoll to stacking](https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/pull/880/commits/54e8f3fc6a7d815325150c4a5b906bd5808a2d93)

Target is to collect bugs about the modifier system and fix them here. 
Might as well throw away some stuff later and do it again with better code. This is more or less a working prototype.

Known Issues:

- [x]  Dropped in Feats/Items that have modifiers math function calls need to have all their modifiers saved by hand to get them working correctly.
- [x] Situational and Constant modifiers have problems to calculate stacking ("insight" with "insight" should not work but "insight" with "enhancement" should and so on) against each other. The result tends to be 0 if the situational modifier is higher and is not able to select the situational modifier if it's lower.
- [x] LookupRange modifiers always behave as if they were their maximum value in bonus stacking calculation
- [x] add all modifier to be working with new system


LebombJames:
There are some bugs introduced by asyncing modifier stacking, actor resources throw errors

image

There are undoubtably lots of smaller bugs introduced by this change, so I'm not entirely happy with it, but admitedly I can't think of a better solution than just doing recalculating them in an update hook/method or something.

That said, if we can find all the bugs and ensure there are no regressions, it should be fine. How much work that will take though is unclear.

Dosendusche:
> Situational and Constant modifiers have problems to calculate stacking ("insight" with "insight" should not work but "insight" with "enhancement" should and so on) against each other. The result tends to be 0 if the situational modifier is higher and is not able to select the situational modifier if it's lower.

I believe everyone knows why this happens but I want to write it down here to have it carved into stone. This is the same on dev and this branch.
Inside every "calculate-[...]-modifier.js" exists a method called "addModifier" and they all have this code:
```
if (bonus.modifierType === SFRPGModifierType.FORMULA) {
                if (item.rolledMods) {
                    item.rolledMods.push({mod: bonus.modifier, bonus: bonus});
                } else {
                    item.rolledMods = [{mod: bonus.modifier, bonus: bonus}];
                }
                return 0;
            }
```
This is done AFTER the stackmodifiers (which does not give a damn if a modifier is situational or not) is executed. This means: Every time stackmodifiers calculates that the situational modifier is the highest mod it just sets the mod to 0.

Fixing this will take some time because I need to figure out how to get out of this dead end.

LebombJames:
Be sure to undo [8bda768](https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/pull/880/commits/8bda7681a0762393efb55873eb0ef316537d5472) and [3aeba30](https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/pull/880/commits/3aeba3012e6a4f60af825f153e30cceb5bbd6b2a) too

LebombJames:
-Situational Modifiers to specific skills appear on all skills
-Default attack modifiers (charge, flanking etc) throw an error since `container.actorId` is undefined

Dosendusche:
> 

what a shame i didn't test this. working on it

LebombJames:
A few oddities I discovered. Just noting them here, I will (try) to investigate them further later:
- Only the highest situational modifier is shown on the roll dialog. Ideally all should be shown, and only one is able to be enabled.
- Didn't dice in constant formulas used to throw an error? It doesn't currently, constant dice just silently get ignored in rolls.

Just FYI too. I think we're gonna do a release soon, without this merged. After that, we can merge this pretty soon after and have it sit on dev for a while so we can find all the bugs.

LebombJames:
Spell DCs have gone missing 😱
image

Dosendusche:
> Spell DCs have gone missing 😱 image

Fixed. Note that we shall never touch that part again or we have to decide to do some major rework.

LebombJames:
Fuck it, let's see what blows up

PR #881, opened 2022-12-23
Add JSON Sanitisation to cook, and improve unpack

LebombJames:
**This merge adds JQuery and JSDOM as prerequisites.**

- Running `unpackPacks` as an NPM script now unpacks the db files directly from the user's foundry data path
- `unpackPacks` and `cook` now run `sanitizeJSON`, which performs the following:
  - Removes  `sort`, `folder`, `effects`, `ownership`/`permissions`, `_stats` and non-essential `flags`,
  - Trims whitespace from item names
  - Replaces Forge URLs in images
  - Strips out redundant AoN/PDF ids and classes (not entirely working currently, why this is drafted)
  - Fixes fake links, replacing them with UUID links (and changed cook to look for UUID links)
  - Runs what was previously `ensureTokenDefaults`
    - Now automatically sets token size according to the actor (we had *a lot* of large+ actors still with 1x1 prototype tokens)
  - A slew of regex to standardise item descriptions

Data pruning can certainly go further, to the point of removing essentially all derived characteristics, as they will be pulled from template.json, or just calculated on actor construction

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -36,6 +36,7 @@
     "eqeqeq": ["error", "always"],
     "no-eq-null": "error",
     "init-declarations": "off",
+    "brace-style": ["warn", "1tbs", { "allowSingleLine": true }],
     "no-undef": "off",
     "no-var": "warn",
     "spaced-comment": "error",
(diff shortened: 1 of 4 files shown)


LebombJames:
Got this 98% working, only things I'm not entirely happy with are that some extrenuous `<div>`s aren't being stripped out, and also we have some random chinese characters in descriptions that can't reliably be replaced. Will have one more look in the future but if nothing comes this'll be merged.

(Prepare for a massive recook after 😛)

Issue #922, opened 2023-01-12
Doshakari series weapons Incorrect damage formula


TheGreatRedDuke:
**What item is incorrect/missing?**
All the Doshakari series weapons (advanced, molecular rift, tactical, ultrathin, and zero-edge) have a misconfigured damage formula I believe.

**What book is the item from? What page is it on? If possible, include an Archives of Nethys link**
Near Space pg. 149 | [Doshakari, Tactical](https://aonsrd.com/WeaponDisplay.aspx?ItemName=Tactical&Family=Doshakari)

**SFRPG system version**
0.22.1

**How is it incorrect?**
The damage formulas have the damage dice, but are all missing the " + @abilities.str.mod". They are fairly standard advanced melee weapons and I can't find anything about them that indicates they shouldn't have the strength mod applied.

**What correction should be made?**
Damage formulas for all these should be "[damage dice] + @abilities.str.mod".

Iankid:
A lot of older items seem to be missing this. Thanks for catching it. Running across more, let us know.

PR #923, opened 2023-01-12
Updates & fixes

Iankid:
Updated Weapon item for later updates. same update in Augmentation PR
Fixed description of Dross guns and Void hide.
Added a few missing missing class features.
Added missing str mod bonus to Doshakari weapons. #922
replaced all mod references in melee damage of melee weapons to abilities.str.mod
removed + bonus reference in damage formulas of weapons


src/items/class-features/improved_terrifying_blast_(ex).json
```diff
@@ -0,0 +1,115 @@
...
+      "sonic": false,
+      "summoning": false,
+      "teleportation": false,
+      "water": false
+    },
+    "duration": {
+      "units": "",
+      "value": null
+    },
+    "formula": "1d4 + 1",
```
LebombJames, line 84: Haven't checked in Foundry, but what does this formula field correspond to on the item?

Iankid: Rounds. Debated adding it there, so your call.

LebombJames: Hmm, not a fan tbh. Better off just doing an inline roll in the item description `[[/r 1d4+1]]{1d4+1 rounds}`

PR #937, opened 2023-01-19
Adding Theme Features / Crisis Themes

TerallianHub:
Added Crisis Refugee Theme
Added Theme Features Compendium


LebombJames:
I'm not a fan of including the Theme's name in the title of the Theme Feature. It should only be the Feature's name. E.g "Care for Others"

TerallianHub:
> I'm not a fan of including the Theme's name in the title of the Theme Feature. It should only be the Feature's name. E.g "Care for Others" (I suppose Theme Knowledge can be exempt of this, so we don't have loads of features all called Theme Knowledge)

I'll keep it just for Theme Knowledge then and remove it from the rest. Should we maybe format it like Weapon Specialization, ie: "Theme Knowledge (Theme Name)"?

LebombJames:
That's a good idea, yeah

TerallianHub:
Latest push should have: Crisis Refugee, Drift Crashed, Opportunist, Planar Traveler, Spectra Scion and Void Nomad, as well as their Theme Features.

LebombJames:
Git seems to think you've renamed and modified some files when creating new ones. Can you verify the files that it thinks you have renamed still exist? Might be worth running `cook` just to refresh your JSONs

TerallianHub:
> 

I've given them another look over and it should be all correct from what I can tell. I did notice in running the cook I'd missed two description links so included those now.

ian612:
I've made a few updates to this to get it ready for merging. I'm going to re-review the changed items to make sure everything's good, and then merge.

src/packs/theme-features.db
danimrath, line 1: Is this supposed to be empty?

ian612 (approved):

PR #946, opened 2023-01-31
level up button for classes

Dosendusche:
I did a thing. Looks like this:
lvlupbuddn
Only shown if you have enough XP.
You press it -> the according class levels up.

### DIFF
--- a/src/lang/de.json
+++ b/src/lang/de.json
@@ -270,7 +270,7 @@
                     "Edit": "Gegenstand bearbeiten",
                     "Equip": "Gegenstand ausrüsten",
                     "Unequip": "Gegenstand ablegen",
-                    "LevelUp": "level up"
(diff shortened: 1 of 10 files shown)


src/lang/de.json
```diff
@@ -269,7 +269,8 @@
                     "Delete": "Gegenstand entfernen",
                     "Edit": "Gegenstand bearbeiten",
                     "Equip": "Gegenstand ausrüsten",
-                    "Unequip": "Gegenstand ablegen"
+                    "Unequip": "Gegenstand ablegen",
+                    "LevelUp": "aufsteigen"
```
LebombJames, line 273: Very minor thing: The level up strings should be capitalised

LebombJames (changes requested):

PR #962, opened 2023-02-10
Quickly change for Quick Change

Iankid:
Quickly added this Change

### DIFF
--- a/src/items/spells/quick_change_(galactic_magic).json
+++ b/src/items/spells/quick_change_(galactic_magic).json
@@ -2,9 +2,6 @@
   "_id": "UeRVCrmlvlwl2HT3",
   "name": "Quick Change (Galactic Magic)",
   "type": "spell",
-  "flags": {
-    "core": {}
-  },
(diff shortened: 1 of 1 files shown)


src/items/spells/quick_change_(galactic_magic).json
```diff
@@ -0,0 +1,135 @@
+{
+  "_id": "UeRVCrmlvlwl2HT3",
+  "name": "Quick Change (Galactic Magic)",
+  "type": "spell",
+  "flags": {
+    "core": {}
+  },
```
LebombJames, line 7:
```suggestion
```
I've clearly broken the flag shaking in cook, so can you remove these?

LebombJames (changes requested):

src/items/spells/quick_change_(galactic_magic).json
LebombJames: Nvm, guess github lets you commit removals from suggestions now

PR #975, opened 2023-02-16
Show weapon attack and damage values on buttons

LebombJames:
image
Contingent on #880's sidebar dragging modifier fixes. Could backport I suppose?


src/module/actor/sheet/base.js
```diff
@@ -395,6 +396,89 @@ export class ActorSheetSFRPG extends ActorSheet {
...
+                if (!proficient) {
+                    parts.push(`-4[${game.i18n.localize("SFRPG.Items.NotProficient")}]`);
+                }
+            }
+
+            const formula = parts.join("+");
+
+            const appropriateMods = item.getAppropriateAttackModifiers(isWeapon);
+            const stackModifiers = new StackModifiers();
+            let modifiers = stackModifiers.process(appropriateMods, null);
```
Dosendusche, line 429: If you want to have this working with #880 please make sure that you do NOT include formula modifiers as they will throw errors when pushed to the non async `stackmodifiers.process()` or at least will not work. Same applies for `_prepareDamageString()` or any other usage of `stackmodifiers.process()`.

LebombJames: `getAppropriateAttackModifiers` and `getAppropriateDamageModifiers` only return constant modifiers, not situational modifiers. They can include things like `@details.level.value`. Will that be alright?

PR #979, opened 2023-02-21
Add Android Links and Modifiers to ArmorUpgrades

rmorgens:
Add links to Android, XenometricAndroid and drone_technomancy
Add modifiers to Electrostatic Fields and Thermal Capacitors


rmorgens:
This replaced my old PR and hopefully is cleaned up and without db files

LebombJames:
Any mention of "race" or "racial" in the mechanical text of the android item (like "alternate racial traits", not any lore/flavor text) should be replaced with "species". This was done globally for the entire system, but it seems those changes aren't on your branch (it's probably easier if you just fix that manually, you've fought enough with git as it is 😄)

LebombJames:
You've definitely still got a few "racial bonus"es and such

rmorgens:
Probably forgot to save one of the files we editted before committing.  I'll take another look and make sure it is cleaned up

PR #995, opened 2023-03-02
[WIP] Timed effects

Dosendusche:
target is to create a new item type that gets deactivated automatically after a certain amount of time.

Tasks:

- [ ] add effect item type
- [ ] add effect item to actor conditions
- [ ] add effect item toggle
- [ ] add effect items to top level "effect" object of the according actor on Actor Update
- [ ] check effect items on "updateWorldTime" if they still should be and if not deactivate them


src/templates/items/effect.hbs
```diff
@@ -0,0 +1,55 @@
...
+    {{!-- Item Sheet Body --}}
+    <section class="sheet-body">
+
+        {{!-- Description Tab --}}
+        {{> "systems/sfrpg/templates/items/parts/item-description.hbs"}}
+
+        {{!-- Details Tab --}}
+        <div class="tab details" data-group="primary" data-tab="details">
+
+            <h3 class="form-header noborder">{{ localize "SFRPG.Items.Feat.FeatureUsage" }}</h3>
```
LebombJames, line 46: Should probably create an `EffectDetails` header: "Effect Details" or something

src/templates/items/parts/item-duration.hbs
```diff
@@ -0,0 +1,29 @@
+<div class="form-group input-select">
```
LebombJames, line 1: The details tab could do with a checkbox to enable/disable the effect itself

src/templates/items/parts/item-duration.hbs
```diff
@@ -0,0 +1,29 @@
...
+            <option value="{{key}}">{{localize name}}</option>
+            {{/each}}
+            {{/select}}
+        </select>
+    </div>
+</div>
+<div class="form-group input-select">
+    <label>{{localize "SFRPG.EffectEndTypesLabel"}}</label>
+    <div class="form-fields">
+        <input type="text" name="system.sourceActorId" value="{{itemData.sourceActorId}}" data-dtype="String" placeholder="{{actor.id}}"/>
```
LebombJames, line 18: This could do with some kind of tooltip, as it is quite an advanced feature.
I also think it's worth adding a dropdown to simplify picking the source actor a bit. It could either have two options "Parent actor" or "other (insert ID)", or alternatively auto-populate from combatants in the current combat (defaulting to just "Parent actor" (or maybe a list of all PCs) when there is no combat

src/templates/items/parts/item-duration.hbs
```diff
@@ -0,0 +1,29 @@
...
+            <option value=""></option>
+            {{#each config.effectDurationTypes as |name key|}}
+            <option value="{{key}}">{{localize name}}</option>
+            {{/each}}
+            {{/select}}
+        </select>
+    </div>
+</div>
+<div class="form-group input-select">
+    <label>{{localize "SFRPG.EffectEndTypesLabel"}}</label>
```
LebombJames, line 16: I prefer something like "Ends on" for the label (which would require "on turn start" to become "turn start")

src/templates/items/parts/item-duration.hbs
```diff
@@ -0,0 +1,29 @@
+<div class="form-group input-select">
+    <label>{{localize "SFRPG.EffectDurationTypesLabel"}}</label>
```
LebombJames, line 2: It would also be neat to include a checkbox for whether the effect's icon should be shown on the actor's token. See `token.toggleEffect`

LebombJames (changes requested):
There's no good place to put this, but having the effect sheet show "Enabled"/"Disabled" beneath the "Effect" header would be neat. See `_getItemStatus` in `item/sheet.js`

LebombJames:
Also somewhere on the sheet a "remaining duration" field would be great; just a read-only bit of text that counted down

src/module/timedEffect/timedEffect.js
```diff
@@ -108,19 +107,22 @@ export default class SFRPGTimedEffect {
             actor.system.timedEffects.find(tEffect => tEffect.id === this.id)?.update(this);
             game.sfrpg.timedEffects.find(gEffect => gEffect.id === this.id)?.update(this);
 
-            actor.update({'items': items});
+            actor.updateEmbeddedDocuments({'Item': effect});
```
LebombJames, line 110: The arguments for `updateEmbeddedDocuments` are different to `update`. They're:
```ts
updateEmbeddedDocuments(embeddedDocumentType, diff[])
```
`embeddedDocumentType` is `"Item"` in this case.
`diff[]` is an array of objects, each object representing changes to an embedded document, and containing the item's `_id`. So you'll end up with something like 
```js
updateEmbeddedDocuments("Item", [
{
  _id: "123456...", 
  system: {
    someProperty: 1000
  }
}
])
```
You can pass full items into the diff array, and Foundry will calculate and apply the diff based on that new item, but creating the diff object yourself will save it from having to do that, and likely be more performant

LebombJames:
I changed `SFRPG.effectDurationTypes` and `effectDurationFrom` from plurals to singulars (`rounds` -> `round`). Mainly because I accidentally migrated activation details to singular forms and we want to keep uniformity between them, but also because singulars are also technically more correct anyway

PR #1076, opened 2023-04-19
Combat Difficulty Calculation

ian612:
This adds functionality to the sfrpg core system to add a combat encounter difficulty display to the combat tab. All calculations are done based on the tokens added to the current active combat. Player-owned PCs are players, Hostile NPCs are enemies.

### Calculated values
- Average party level (of player-owned PC tokens)
- Encounter enemy total XP value
- Encounter Challenge Rating
- Encounter difficulty (https://aonsrd.com/Rules.aspx?ID=228)
- Challenge Rating of the highest CR enemy that can still be added to the encounter without changing the difficulty
- Per Player XP value for completing the encounter
- Suggested wealth gain for the encounter (https://aonsrd.com/Rules.aspx?ID=230)

#### Starship Combat Calculations
- Player ship(s) effective tier
- Enemy ship(s) effective tier
- Difficulty of combat based on tier comparison
- Combat total XP value
- Suggested wealth gain for the encounter

### UI Features
- Display color coding for difficulty (blue for easy, green for average, yellow for challenging, etc.)
- By clicking the difficulty, a popout is rendered that includes detailed statistics about the encounter
- A setting was added to enable or disable the display


src/module/combat/combat.js
```diff
@@ -552,6 +552,153 @@ export class CombatSFRPG extends Combat {
...
     }
 
+    average(array) {
+        let total = 0;
+        let count = 0;
+        jQuery.each(array, function(index, value) {
+            total += value;
+            count++;
+        });
+        return total / count;
```
LebombJames, line 562: I assure you you don't need jQuery for this 😄
```js
const average = (array) => array.reduce((total, value) => total + value), 0) / array.length;
```
We also don't need to add this function to the class itself, it can be declared inside the `calculateAPL` function to keep it local

src/module/combat/combat.js
```diff
@@ -552,6 +552,153 @@ export class CombatSFRPG extends Combat {
...
+    }
+
+    getEncounterInfo() {
+        if (!this.flags.sfrpg) {
+            this.flags.sfrpg = [];
+        }
+
+        const playerInfo = this.calculateAPL();
```
LebombJames, line 570: We can make these array indices look a bit more ergonomic by destructuring them
```js
const [PCs, APL] = this.calculateAPL();
this.flags.sfrpg.PCs = PCs;
this.flags.sfrpg.APL = APL;
```
for example

src/module/combat/combat.js
```diff
@@ -552,6 +552,153 @@ export class CombatSFRPG extends Combat {
...
+        const XPTable = CONFIG.SFRPG.XPTable;
+        const APL = this.flags.sfrpg.APL;
+        const numPlayers = this.flags.sfrpg.PCs.length;
+        const numEnemies = this.flags.sfrpg.enemies.length;
+        const XPtotal = this.flags.sfrpg.enemyXP;
+        const XPNoCR = {CR: 0, minXP: 0, totalXP: 0, indOneThree: 0, indFourFive: 0, indSixPlus: 0};
+
+        // Check that Players and NPCs are both present
+        if (!numPlayers) {
+            return [XPNoCR, `${game.i18n.format("SFRPG.Combat.Difficulty.Levels.NoPCs")}`, 0];
```
LebombJames, line 641: These strings don't need to be template literals, you can just call `game.i18n.format`

src/module/config.js
```diff
@@ -1400,6 +1400,39 @@ SFRPG.CR_EXP_LEVELS = [
     1638400
 ];
 
+SFRPG.XPTable  = [
```
LebombJames, line 1403: This would be better as an object of objects, with the key as the CR value.
```js
SFRPG.XPTable  = {
  "1/8": {minXP: 0, totalXP: 50, indOneThree: 15, indFourFive: 15, indSixPlus: 10},
  "1/6": {minXP: 50, totalXP: 65, indOneThree: 20, indFourFive: 15, indSixPlus: 10},
}
```
Alternatively for easier key, value lookup, and the ability to store numbers as keys, you could use a Map. Also what does `ind` stand for?

LebombJames (commented):
Just a few things I noticed 🙂

src/module/combat/combat.js
ian612: Cool, yeah this seems a lot simpler. Implemented.

src/module/combat/combat.js
ian612: This makes sense, I didn't know you could assign like this in JS.

Would it make more sense to just directly assign to the flags, or would this be bad practice?

```suggestion
        [this.flags.sfrpg.PCs, this.flags.sfrpg.APL] = this.calculateAPL();
```

src/module/combat/combat.js
ian612: Fixed

src/module/config.js
ian612: ind was standing in for individual, so indOneThree was individual XP award per player when 1-3 players are in the combat. indFourFive and indSixPlus were the same. I can probably update them to something more descriptive like perPlayerXP123, perPlayerXP45, perPlayerXP6?

src/module/combat/combat.js
LebombJames: I'm not sure if that even works tbh, but even if it does I prefer `[PCs, APL]` be sure it provides more information about what is coming out of the function

src/module/config.js
LebombJames: `perPlayerXP` works well

src/module/combat/combat.js
```diff
@@ -552,6 +552,159 @@ export class CombatSFRPG extends Combat {
         return this.getIndexOfFirstUndefeatedCombatant() === null;
     }
 
+    getEncounterInfo() {
+        if (!this.flags.sfrpg) {
+            this.flags.sfrpg = [];
```
LebombJames, line 557: Just spotted this too; `this.flags.sfrpg` should be an object

src/module/config.js
```diff
@@ -1400,6 +1400,51 @@ SFRPG.CR_EXP_LEVELS = [
...
+    "easy": "SFRPG.Combat.Difficulty.Levels.Easy",
+    "average": "SFRPG.Combat.Difficulty.Levels.Average",
+    "challenging": "SFRPG.Combat.Difficulty.Levels.Challenging",
+    "hard": "SFRPG.Combat.Difficulty.Levels.Hard",
+    "epic": "SFRPG.Combat.Difficulty.Levels.Epic",
+    "greater-than-epic": "SFRPG.Combat.Difficulty.Levels.GreaterThanEpic"
+};
+
+SFRPG.CRMap = new Map();
+SFRPG.CRMap.set("0", {minXP: 0, totalXP: 0, perPlayerXP123: 0, perPlayerXP45: 0, perPlayerXP6: 0});
```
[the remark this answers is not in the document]
ian612: How does that work with fractions? If I set a key as 1/4 I assume it'll convert to 0.25, and I'll have to convert that back to a string for display later?

src/module/combat/combat.js
ian612: Fixed

src/module/config.js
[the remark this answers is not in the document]
LebombJames: That's true. Realistically I just want to avoid using `eval`, perhaps you could parse `"1/2"` into`0.5` by splitting the fraction by the `/`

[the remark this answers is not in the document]
LebombJames: If you don't wanna do numbers, then there's no real reason to use the Map, you can lookup objects just fine with `CRMap["1/2"]`; up to you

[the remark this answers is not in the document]
ian612: Okay yeah I just swapped it over to using an object then, I think it'll be a bit simpler this way rather than doing double conversions for the fractional values.

LebombJames:
Namespacing the CSS using LESS's nesting would ensure it doesn't mess with anything else

```less
.encounter-difficulty-sheet {
  .encounter-table-header { /* ... */ }
  ul { 
    li { /* ... */ }
   }
}
```
You can do the same for the difficulty display on the combat tracker

ian612:
This is now complete, should require no further changes on my end to get all the features I wanted to include working.

ian612:
Found a bug in the leftover CR calculation, had swapped a < for a <= and a >= for a >.

Old code
image

New code
image

LebombJames:
Having thought about this a bit further, I think the window would be better as its own Application, rather than a Dialog, so it's not swimming in the same pool as the CombatSFRPG class. Should be as simple as setting up a new class (possibly accepting the current combat as a parameter to the constructor) and moving all the related methods out of CombatSFRPG and into it.

ian612:
> Having thought about this a bit further, I think the window would be better as its own Application, rather than a Dialog, so it's not swimming in the same pool as the CombatSFRPG class. Should be as simple as setting up a new class (possibly accepting the current combat as a parameter to the constructor) and moving all the related methods out of CombatSFRPG and into it.

Cool, I've set it up in it's own app. I think I've set it up correctly, but feel free to make tweaks as you see fit.

src/module/combat/combat.js
```diff
@@ -1192,11 +775,11 @@ Hooks.on('renderCombatTracker', (app, html, data) => {
...
             console.log("Starfinder | Rendering Encounter Statistics Dialog");
             const isStarship = activeCombat.flags.sfrpg.combatType === "starship";
-            const contentTemplate = `//systems/sfrpg/templates/combat/${isStarship ? "starship" : "normal"}-encounter-stats.hbs`;
+            const contentTemplate = `//systems/sfrpg/templates/apps/${isStarship ? "starship" : "normal"}-encounter-stats.hbs`;
 
             new Dialog({
-                title: `${game.i18n.format("SFRPG.Combat.Difficulty.Tooltip.Details")}: ${CONFIG.SFRPG.difficultyLevels[activeCombat.flags.sfrpg.difficulty]} ${game.i18n.format("SFRPG.Combat.Difficulty.Tooltip.DifficultyEncounter")}`,
-                content: await renderTemplate(contentTemplate, activeCombat),
+                title: `${game.i18n.format("SFRPG.Combat.Difficulty.Tooltip.Details")}: ${CONFIG.SFRPG.difficultyLevels[diffObject.difficultyData.difficulty]} ${game.i18n.format("SFRPG.Combat.Difficulty.Tooltip.DifficultyEncounter")}`,
+                content: await renderTemplate(contentTemplate, diffObject),
```
LebombJames, line 782: Nearly there, the point of the Application is to render it *with* the Application, not just use an instance of a class as the Dialog data. 
You'll want to instantiate a CombatDifficulty here and `.render(true)` it (or just render it if you need to instantiate it earlier). Then in CombatDifficulty, you'll want to create the data object used in the handlebars in the `getData` method. You can also create getters for the `title` and `template`

LebombJames (changes requested):

src/module/apps/combat-difficulty.js
```diff
@@ -0,0 +1,434 @@
+/**
+ * A class for checking the difficulty of the currently defined combat
+ *
+ * @type {Application}
+ */
+export class CombatDifficulty extends Application {
+    constructor(combatData, options = {}) {
+        super(combatData, options);
+        this.options.classes = ["sfrpg", "dialog"];
```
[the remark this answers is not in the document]
LebombJames: `app` is actually the correct class 😛

src/module/combat/combat.js
LebombJames: In that case just instantiate earlier, then render it here. You can keep the methods to add the button display to the combat tracker on the CombatSFRPG class itself

LebombJames:
Do you plan to implement Vehicle Chase XP, or should the difficulty display be hidden for Vehicle Chases?

src/module/combat/combat.js
ian612: Fixed in
https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/pull/1076/commits/8d9790b320d4d213c413d60bcec61feb11a4da2e

ian612:
> Do you plan to implement Vehicle Chase XP, or should the difficulty display be hidden for Vehicle Chases?

My intent was to have it be hidden. I've never used vehicle chases before so I'm not that familiar with them and how they're implemented in the system, so haven't looked into how difficult figuring out the XP would be. If it's not too difficult I could implement it.

ian612:
I've noticed a bug where if 2 or more tokens that are in the active combat are deleted simultaneously, I get an error though everything seems to work fine.

image

Weirdly, adding a console.log(combatant) in parseShips() to view the combatant data structure that's causing the error causes the error to go away...

ian612:
> I've noticed a bug where if 2 or more tokens that are in the active combat are deleted simultaneously, I get an error though everything seems to work fine.
> 
> image
> 
> Weirdly, adding a console.log(combatant) in parseShips() to view the combatant data structure that's causing the error causes the error to go away...

Fixed it in https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/pull/1076/commits/1eb4b076433f607341e52423b4509686432b329e by adding optional chaining to checks of token disposition.

ian612:
I moved the code in combat.js outside of the if/else statement so that
a) it only calculates for GMs, since players should never need to see this window or access this information
b) the difficulty indicator persists after the combat has started, so that the difficulty can still be seen and the encounter stats accessed (such as if you want to check how much XP players get after the combat is finished)

LebombJames:
I make this about done. Anything else you wanna add?

ian612:
Yeah I’m happy with it as well. I’m good to close it out and merge if you are.

Issue #144, opened 2020-08-20
Weapon View should list only existing properties when in view mode


Juxtaposedwords:
When viewing a weapon (and not in edit mode) we should see only the viewed properties. So when viewing "Weapon Details" under "Weapon Properties" for the Advanced Autobeam Artillery you would see only the applicable terms (Which would ideally have a mouse linking to or explaining the term):

Advanced Autobeam Artillery:
**Weapon Type:** Heavy Weapon
**Weapon Properties:** Two-handed, Automatic

wildj79:
![](https://github.trello.services/images/mini-trello-icon.png) [[FEAT] Weapon View should list only existing properties when in view mode](https://trello.com/c/kVIGv6U2/38-feat-weapon-view-should-list-only-existing-properties-when-in-view-mode)

Imper1um:
I think this is a little old, and was never closed properly.

ian612:
You are correct, this was implemented at some point and is present as of 0.27.0

PR #1145, opened 2023-06-07
Trait selector update

ian612:
This PR updates the trait selector app.

### New/Modified Features List
- [x] Restrict the maximum height of the trait selector
- [x] Draw traits in 3 columns instead of one
- [x] Add a text search/filter field
- [ ] Set up weapon properties to use this trait selector
- [ ] Update visual styling to look nicer

### Stretch Goal
- [ ] Set selected (checked) weapon properties to include a text field that can be edited (e.g. to indicate something like Explode (5ft))
- [ ] Backend data structure updates to enable this
- [ ] Migration script to not kill everyone's world actors

### DIFF
--- a/src/module/apps/trait-selector.js
+++ b/src/module/apps/trait-selector.js
@@ -9,10 +9,11 @@ export class TraitSelectorSFRPG extends FormApplication {
 
         options.id = "trait-selector";
         options.classes = ["sfrpg"];
-        options.title = "Actor Trait Selection";
+        options.title = "Trait Selection";
         options.template = "systems/sfrpg/templates/apps/trait-selector.hbs";
-        options.width = 320;
+        options.width = 480;
         options.height = "auto";
+        console.log(options);
 
         return options;
     }
@@ -32,20 +33,20 @@ export class TraitSelectorSFRPG extends FormApplication {
      * @returns {Object}
      */
     getData() {
-        let attr = getProperty(this.object, this.attribute);
+        const attr = getProperty(this.object, this.attribute);
         if (typeof attr.value === "string") attr.value = this.constructor._backCompat(attr.value, this.options.choices);
 
         const choices = duplicate(this.options.choices);
         const isEnergyResistance = this.attribute === "data.traits.dr";
         if (!isEnergyResistance) {
-            for (let [k, v] of Object.entries(choices)) {
+            for (const [k, v] of Object.entries(choices)) {
                 choices[k] = {
                     label: v,
                     chosen: attr.value.includes(k)
                 };
             }
         } else {
-            for (let [k, v] of Object.entries(choices)) {
+            for (const [k, v] of Object.entries(choices)) {
                 choices[k] = {
                     label: v,
                     chosen: false,
@@ -80,7 +81,7 @@ export class TraitSelectorSFRPG extends FormApplication {
         if (!current || current.length === 0) return [];
         current = current.split(/[\s,]/).filter(t => !!t);
         return current.map(val => {
-            for (let [k, v] of Object.entries(choices)) {
+            for (const [k, v] of Object.entries(choices)) {
                 if (val === v) return k;
             }
             return null;
@@ -98,13 +99,13 @@ export class TraitSelectorSFRPG extends FormApplication {
         let choices = [];
 
         if (this.attribute !== "system.traits.dr") {
-            for (let [k, v] of Object.entries(formData)) {
+            for (const [k, v] of Object.entries(formData)) {
                 if ((k !== 'custom') && v) choices.push(k);
             }
         } else {
             let resistances = Object.entries(formData).filter(e => e[0].startsWith("er"));
             resistances = resistances.reduce((obj, entry) => {
-                let [type, i] = entry[0].split('.').slice(1);
+                const [type, i] = entry[0].split('.').slice(1);
 
                 if (!obj[type]) obj[type] = {};
                 obj[type][i] = entry[1];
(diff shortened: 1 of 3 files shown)


ian612:
This should also resolve https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/issues/144, ideally

src/module/apps/trait-selector.js
```diff
@@ -9,120 +9,164 @@ export class TraitSelectorSFRPG extends FormApplication {
...
+    getData() {
+        console.log(this);
+
+        // Initialize variables for easy access
+        const dataFormat = this.options.format;
+        const dataLocation = this.options.location;
+        const traitData = getProperty(this.object, dataLocation);
+
+        // Choose the appropriate data parser to create the form
+        switch (dataFormat) {
```
LebombJames, line 34: So I think instead of conditionally returning data to the template depending on `format`, each type of format could correspond to a subclass of `TraitSelectorSFRPG`, which declares its own methods for getting and getting its trait choices. For example, the actor trait selector in base.js would instantiate an `ActorTraitSelectorSFRPG`.

Then you could end up with something like this in a "base trait selector`
```js
getData() {
        console.log(this);

        // Initialize variables for easy access
        const dataFormat = this.options.format;
        const dataLocation = this.options.location;
        const traitData = getProperty(this.object, dataLocation);

       return this._getTraitChoices(traitData);
```
The base trait selector doesn't define a `_getTraitChoices`, but each subclass would, so it would call that on the subclass (may require a `super.getData()`).

ian612: Alright I think I've changed it in the way you suggested. Now there's a subclass for actor traits that extends the base trait selector, and the switch statements are gone. Let me know if that looks better.

ian612: This does mean that for each sheet type (actor, item) where the trait selector is being called, only that one subclass can be called and there is an assumed data format that can't differ between traits on that sheet.

Unless, of course, you added a switch on the `dataFormat` property within the subclass, which I guess is just how you'd do it if it came to that.

LebombJames: Looks good. Yes, the intent is if a single sheet calls multiple types it uses something like this to determine which:
```js
const cls = {
  "actorTraits": ActorTraitSelector,
  "someOtherTrait": SomeOtherTraitSelector
}[options.format];
new cls(options);
```

ian612: Cool, didn't know you could write it that way. I'll carry on applying it to weapon properties, then.

ian612:
Alright I think this is ready for review. Everything is working, looks nice, and the migration script should hopefully not explode anything.

For now I've added a check in `migration.js` (called `performMigrate`) to enable/disable the actual data migration function and left it off by default. Making a note here that this should be removed before actual release.

src/module/apps/trait-selector.js
```diff
@@ -9,120 +9,93 @@ export class TraitSelectorSFRPG extends FormApplication {
...
+        });
 
-        this.object.update({
-            [`${this.attribute}.value`]: choices,
-            [`${this.attribute}.custom`]: formData.custom
+        // re-render the template with updated data if a trait is selected/unselected
+        html.on('change', 'input[type=checkbox]', async ev => {
+            const formData = this._getSubmitData();
+            await this._updateObject(ev, formData);
+            this.render();
```
LebombJames, line 97: We shouldn't be updating the actor as soon as checkboxes are ticked, we should wait till the form is submitted. I understand why you're doing this, so you can just re-render it, but another solution should be used. Best I can suggest is a bunch of event listeners shuffling things into place as needed,

src/module/apps/trait-selectors/weapon-property-selector.js
```diff
@@ -0,0 +1,75 @@
+import { TraitSelectorSFRPG } from "../trait-selector.js";
+
+export class WeaponPropertySelectorSFRPG extends TraitSelectorSFRPG {
+    constructor(item, options) {
+        super(item, options);
+        super.getData();
```
LebombJames, line 6: Don't need to getData here, it's called during `render` anyway

LebombJames (changes requested):

src/module/apps/trait-selectors/weapon-property-selector.js
ian612: Removed

src/module/apps/trait-selector.js
ian612: I've updated this now to shuffle traits around using event listeners instead of always submitting

ian612:
This is now fixed up and requested changes have been made. Should be ready for review. Note that I've turned migration off for now, but when weapon properties get updated they store in the new format. It's backward compatible so this shouldn't be an issue as long as this branch is being used, but if you review this branch, make changes to an item's weapon properties, then go back to another branch, those properties likely won't be read correctly by the old trait selector.

ian612:
I believe this should be good to add too the system. Might as well soon before I have to rewrite it into appV2  XD

LebombJames:
I hate to delay this even further, but I think this would be a great candidate for use with DataModels, and I think doing a huge migration now (when we may need to do another with DataModels) would be a poor idea. The DataModel conversion is the very first thing I want to do after V13 so hopefully we can get back to this real soon

ian612:
All good, no worries. I was joking in my previous comment but as I think about it more, it may honestly make sense to take a crack at rewriting this in AppV2 with DataModels to get used to how they work (and leave out the bits that need migration for now to simplify it).

ian612:
This has now been updated to use DataModels, and to clean it up a little for v13 compatibility.

src/module/item/item.js
```diff
@@ -874,25 +874,27 @@ export class ItemSFRPG extends Mix(foundry.documents.Item).with(ItemActivationMi
...
+                abl = "";
+            } else if (this.type === "spell") {
+                if (itemData.actionType === "rsak") {
+                    abl = "dex";
+                } else if (itemData.actionType === "msak") {
+                    abl = "str";
+                } else {
+                    abl = actorData.attributes.spellcasting || "int";
+                }
+            } else if (itemData.properties?.operative?.value && actorData.abilities.dex.value > actorData.abilities.str.value) {
```
danimrath: Same comment as above.

src/module/actor/sheet/base.js
```diff
@@ -456,25 +458,27 @@ export class ActorSheetSFRPG extends foundry.appv1.sheets.ActorSheet {
...
+                    abl = "";
+                } else if (this.type === "spell") {
+                    if (itemData.actionType === "rsak") {
+                        abl = "dex";
+                    } else if (itemData.actionType === "msak") {
+                        abl = "str";
+                    } else {
+                        abl = actorData.attributes.spellcasting || "int";
+                    }
+                } else if (itemData.properties?.operative?.value && actorData.abilities.dex.value > actorData.abilities.str.value) {
```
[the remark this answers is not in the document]
ian612: Yeah should be. Previously there was a check for `!abl` in all of the statements, so I figured it was more performant to just wrap the whole block in it and not check for it every time. The exception was the check for the operative property, which was overridden regardless of whether an ability was specified. That seemed a bit odd to me, so I lumped it in here with the others.

All that said, I thought I did this on a different branch but must have forgotten to switch it... It doesn't really make sense to have it in this one as it's an unrelated change

src/module/item/item.js
ian612: Fixed by reverting back to previous version

danimrath:
Also identified that the change is preventing properties from spells printing to chat. 
image
image

src/module/item/item.js
```diff
@@ -558,7 +558,7 @@ export class ItemSFRPG extends Mix(foundry.documents.Item).with(ItemActivationMi
         }
 
         // Filter properties and return
-        data.properties = props.filter(p => !!p && !!p.name);
+        data.properties = props.filter(p => !!p.value && !!p.name);
```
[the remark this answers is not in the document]
ian612: Good catch. I have no idea why I changed this, `p` should only have `name`, `tooltip`, and `title` as parameter options. I've changed it to just check for the name property, as long as that's not blank it should be included.

danimrath (approved):
Seems to work correctly.

danimrath:
I did have an issue with the items copying over each other on a rare occasion though fyi. Couldn't pinpoint it to this PR.

ian612:
Yeah the commit that fixed that was later than this branch, so it should be fine once this is merged into dev

PR #1152, opened 2023-06-14
Archetype features

Iankid:


danimrath:
Unfortunately this has been sitting for a long time and needs to have dev merged into it and seems to have a significant amount of files needing updates.

ian612:
I've found only minor issues so far in this PR through ~40 items I reviewed, so I think it's fine to include as is. Any other corrections can be made as needed later on a case-by-case basis.

PR #1160, opened 2023-06-16
Iconics

JonDemand:
Updated Iconic Character Sheets to include all automation since they were originally added

Also included new artwork direct from Paizo

and new tokens created from that artwork


LebombJames:
Weren't you updating the license file too?

JonDemand:
Will do that as a seperate PR,  as it's going to take me a while to gather all the credits,  not required for this one

LebombJames:
- Altronus L8 has no damage type set for Solar Weapon
- Some Skill Expertise modifiers on Navasi (saw one on L4, check L8 too) are set to constant; they must be situational.
- You may want to set a fixed DC for Explosive Blast on Raia L8 since the DC is calculated lower due to the fact its KAS is set to dex for the attack roll.
- Velloro L8 entropic pool max is incorrectly set to 3, not 6.

JonDemand:
updated those errors,  have to be especially careful with these,  as it needs to match the PDF,  even if that conflicts with the system rules,  but these were just errors on my part

LebombJames:
Why did that commit have several thousand lines changed for each file?

LebombJames:
Nvm, you just didn't cook and everything got spewed out in a random order. That's fine

Issue #1062, opened 2023-04-16
Envoy Improvisations Missing


JonDemand:
**What item is incorrect/missing?**
Brace Yourselves
Improved Brace Yourselves

**What book is the item from? What page is it on? If possible, include an Archives of Nethys link**
Armory
https://www.aonsrd.com/Improvisations.aspx?ItemName=Brace%20Yourselves
https://www.aonsrd.com/Improvisations.aspx?ItemName=Improved%20Brace%20Yourselves

**SFRPG system version**
23.0

PR #992, opened 2023-03-01
Add missing Envoy Improvisations - Armory, AP#25, etc

rmorgens:
Adding missing Envoy Improvisations from Armory, AP#25 and a few other clean-ups related to Envoy Improv's

PR #1174, opened 2023-06-23
added brace yourselves improvisations

ian612:
Adds missing envoy improvisations from #1062


LebombJames:
Any particular reason to add new icons to the system for these?

ian612:
I noticed that most icons were named for the feature/item they're used for. I assumed we wanted to keep it that way so I duplicated them and changed the names to match. I can change it and link them to the icons that already exist if that's the preferred way of doing it.

ian612:
Also I realized this is also open under https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/pull/992, I can close it and let that larger one take precedence

Issue #1052, opened 2023-04-08
Armor upgrade displaying incorrect allowed armor in sheet.


Iankid:
On sheet, the Armor upgrade item displays the allowed armor on it. However it is only displaying Power Armor, no matter what's set in the item.

![image](https://user-images.githubusercontent.com/60759323/230736824-4d419d39-5823-4fe0-875c-a462b498bac1.png)

 - SFRPG Version: Dev Branch

ian612:
This might be a bug with your UI module. I can select any allowed armor type for armor upgrades.

![image](https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/assets/34078802/6bf7e8c2-0d02-4164-be2f-be51bbfb675a)

On the left are jump jets, on the right is a jetpack, I could modify it to any and it shows up properly.
![image](https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/assets/34078802/67efa012-f01e-4921-8f90-b403119d6d71)

ian612:
Ah wait I see, it's not functioning correctly on the _character_ sheet, not the item sheet.
![image](https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/assets/34078802/fb693f70-825a-40af-9071-ab26da43d861)

I'll look into it

PR #1182, opened 2023-06-26
Armor Upgrade Item Summary & Tooltip Fix

ian612:
Fixes issue #1052, where the upgrade chat data was being gathered incorrectly from upgrade items. Also fixes an issue where the properties tooltip was displaying "undefined" for properties without one.


ian612:
Tooltip bug this fixes shown below.

image

LebombJames:
My concern with this is with how it relates to existing upgrades; some have `armorType` set, others `allowedArmorType` and others both. It seems `allowedArmorType` is the one more commonly used

ian612:
What's the resolution for that? Go through and edit all the upgrades in the compendium to use `allowedArmorType`?

LebombJames:
As far as I can tell, on any item with both `armorType` and `allowedArmorType`, `allowedArmorType` is correct. If an item has only `armorType`, then *that* is correct. Therefore I think the way forward is to simply delete `armorType` on items with both, and migrate `armorType` to `allowedArmorType` on items with only `armorType`.

ian612:
> As far as I can tell, on any item with both `armorType` and `allowedArmorType`, `allowedArmorType` is correct. If an item has only `armorType`, then _that_ is correct. Therefore I think the way forward is to simply delete `armorType` on items with both, and migrate `armorType` to `allowedArmorType` on items with only `armorType`. Migrate in the most basic definition of the word; just needs to be a simple compendium node script.

I've added a node script for this in https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/pull/1182/commits/8efad9426ddaf9eeb8799570528f2d26bf10431a and updated the relevant compendium items.

src/module/actor/sheet/base.js
```diff
@@ -1020,8 +1020,9 @@ export class ActorSheetSFRPG extends ActorSheet {
             Hooks.callAll("renderItemSummary", this, div, {}); // Event listeners need to be added to this HTML.
 
             const props = $(`<div class="item-properties"></div>`);
+            console.log(chatData.properties);
             chatData.properties.forEach(p => props.append(
-                `<span class="tag" data-tooltip="${p.tooltip || p.title}"><strong>${p.title ? p.title + ":" : ""} </strong>${p.name}</span>`
+                `<span class="tag" data-tooltip="${(p.tooltip || p.title) ? (p.tooltip || p.title) : ""}"><strong>${p.title ? p.title + ":" : ""} </strong>${p.name}</span>`
```
LebombJames, line 1025: It'd be better to only add a tooltip.at all if there is a tooltip or title, rather than producing an empty tooltip

src/module/actor/sheet/base.js
```diff
@@ -1020,8 +1020,9 @@ export class ActorSheetSFRPG extends ActorSheet {
             Hooks.callAll("renderItemSummary", this, div, {}); // Event listeners need to be added to this HTML.
 
             const props = $(`<div class="item-properties"></div>`);
+            console.log(chatData.properties);
```
LebombJames, line 1023: Errant logging

LebombJames (changes requested):

src/module/actor/sheet/base.js
ian612: Fixed both items as suggested

src/module/actor/sheet/base.js
```diff
@@ -1020,9 +1020,20 @@ export class ActorSheetSFRPG extends ActorSheet {
             Hooks.callAll("renderItemSummary", this, div, {}); // Event listeners need to be added to this HTML.
 
             const props = $(`<div class="item-properties"></div>`);
-            chatData.properties.forEach(p => props.append(
-                `<span class="tag" data-tooltip="${p.tooltip || p.title}"><strong>${p.title ? p.title + ":" : ""} </strong>${p.name}</span>`
-            ));
+            console.log(chatData.properties);
```
LebombJames, line 1023: You added it back! 😆

PR #1185, opened 2023-06-29
Various updates

Iankid:

PR #1204, opened 2023-07-16
Rules updates

Iankid:


LebombJames:
You seem to have mixed in #1185 in this PR.

PR #1211, opened 2023-07-20
De localisation update

Beff42:
Hi there,

this pull request contains mainly a localisation for the German language.
Additionally it fixes two small things:
- It fixes a translation issue by adding a new localisation tag for savingThrow (save and saving throw are the same word in English, not in German though)
- It adds a bit more info to the contributing.md. Namely where to find the foundry.js and a clarification, that the foundryconfig.json should point to the parent path of the data folder, not the data folder itself.

Let me know, if you think I should rather split those into separate pull-requests, or if I need to follow some other guideline for a pull-request.


LebombJames:
I'm fine with splitting `Save` and `SavingThrow`, but I'd rather `Save` be (in English) "Save" and `SavingThrow` be "Saving Throw". Then on item cards `Save` is used. Does that work in German?

Also conflicts!

LebombJames:
Ah, I think I understand the problem better now. Do you mean "Save" as in "Save to disk" is different to "Save" as in "Reflex Save"? If so, I'd rather create a new localisation key for "Save" (to disk)

Beff42:
I did not know that I just could pull the foundry localisation. Nice, that makes the changes a lot smaller!
I'm done with my changes, sorry for the myriad commits ^^'

PR #1221, opened 2023-07-25
Updated Polymorphic weapons

Iankid:


src/items/equipment/polyclub_advanced.json
```diff
@@ -12,26 +12,48 @@
     "actionTarget": "kac",
     "actionType": "mwak",
     "activation": {
-      "type": "none",
+      "type": "action",
```
LebombJames, line 15: Why do these all have activation of action now?

Iankid: Guess this is a point of determining our standard here. At least Ranged weapons currently have to have activation in order to list range. So should we, or should we not, have that set for melee weapons? Either way, we have melee weapons going back and forth on this.
I suppose for ranged, the button on the sheet is hidden. So your call.

Iankid: oh wait, I see. the guns are all none. which doesn't show a button for weapons. alright, I'll change it back.

PR #1240, opened 2023-08-14
Untrained skills update

### LINKED ISSUE #676: [Feature Request] Add "Hide Untrained Skills" toggle
I'd like a way to trim down the character sheet just a little, and I think having the ability to hide skills with zero ranks would do the job. Make it a toggle for those times you have to roll anyway, or for when you're gaining a level and picking up something new.

Imper1um:
resolves #676 
A new option has been added to below the Skills called "Hide Untrained." Checking this option will hide all skills that cannot be used Untrained, if they do not have at least one rank in them.

Here's how it looks in the various ways (I use the Dark UI Module on my personal server).

Hide Untrained
Hide Untrained - Show All One
Hide Untrained - Show None One

A huge infinite loop bug was found while fixing another bug that I found. Basically, the entire calculate folder was calling for the entire calculation stack to be processed, leading to an infinite loop. The entire system was causing a StackOverflowException, being discarded, and then it just... went as normal. However, the entire process was filling up memory if you were in Inspect Element, causing the entire Inspect Element UI to crash. So, I removed the entire stack calling the recalculation of the stack in the various .js files.

### DIFF
--- a/src/lang/en.json
+++ b/src/lang/en.json
@@ -2557,6 +2557,8 @@
         "SkillTooltipThemeMod": "Theme Modifier: {mod}",
         "SkillTooltipTrainedClassSkill": "Trained Class Skill Modifier: {mod}",
         "SkillTrainedOnly": "Trained Only",
+		"SkillTrained": "Trained",
+		"SkillUntrained": "Untrained",
         "SkillTrainedOnlyDialog": {
             "Content": "{skill} is a trained only skill, but {name} is not trained in that skill. Would you like to roll anyway?",
             "Title": "{skill} is trained only"
--- a/src/less/actors.less
+++ b/src/less/actors.less
@@ -427,6 +427,12 @@
     /* ----------------------------------------- */
     /*  Skills                                   */
     /* ----------------------------------------- */
+	
+	/* If hideUntrained is selected, only show skills that are allowed untrained,
+	 * or trained skills that have at least one rank in it. */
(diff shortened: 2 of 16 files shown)


LebombJames:
Can you please separate the hide untrained functionality and the bugfixes into separate PRs? Could you also provide some reproduction steps for triggering the StackOverflow? That seems like a very serious bug but I've never encountered it myself and have had no reports of it.

Sidenote, I also recommend installing ESLint and Editorconfig in your IDE. There are various minor formatting/code style issues that those extensions would handle for you.

Imper1um:
Sure, I'll separate out these fixes into separate pull Requests. Give me a little bit to separate the branches and create new pull requests.

As for the reproduction, its been in the code for a while it looks like. The problem is they got hidden. What was happening is that there was an Exception on calculate-encumbrance.js which I had a serious annoyance about. It was popping up all over when I was running it. I found that on Line 11 of calculate-encumbrance.js there is an issue where encumbrance.tooltip is referenced prior to it being initialized on Line 55. So, I fixed it... and the StackOverflow errors came out by the dozen. All of it was because all of these calculate- js were all referencing stackModifiers.process inside of them. The thing is: Its completely unnecessary. Its like adding to the stack after the stack has been processed.

When I stripped out the .process out of all of the calculate- js, not only did I not see any changes to the calculation, but it killed the StackOverflow issues.

Imper1um:
No longer a valid pull request. This Pull Request has been split into #1241 #1242 #1243

PR #1242, opened 2023-08-14
Fix Recursion Bug in rules

Imper1um:
There was a very serious issue in the rules js. Most of the calculate- js files had this issue, but it was being hidden by a crash occurring in calculate-encumbrance.js . The crash was minor, but it did result in a crash of the encumbrance calculation (no idea why encumbrance was still able to be calculated properly). Fixing the crash resulted in StackOverflowExceptions to bubble up from most of the calculate- .js files.

The problem is that these calculate- .js files were calling stackModifiers.process . However, these calculate- .js files were, in fact, stackModifiers being processed. What was happening is these stackModifiers were being added to the stack infinitely, causing the StackOverflowException because it was going in a loop. So, removing the stackModifiers.process from all the calculate- .js fixes the issue. They weren't necessary anyways; the idea behind the stackModifiers is the process *should* have been run prior to each calculation step being necessary.

### DIFF
--- a/src/module/rules/actions/actor/calculate-ability-check-modifiers.js
+++ b/src/module/rules/actions/actor/calculate-ability-check-modifiers.js
@@ -28,27 +28,26 @@ export default function(engine) {
         const filteredMods = modifiers.filter(mod => {
             return (mod.enabled || mod.modifierType === "formula") && [SFRPGEffectType.ABILITY_CHECK, SFRPGEffectType.ABILITY_CHECKS].includes(mod.effectType);
         });
+		
+		const getFilteredAbilities = (abl, ability, mod) => {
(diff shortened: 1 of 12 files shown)


LebombJames:
With this change bonuses stack unexpectedly. For example, +5 and +10 enhancement bonuses to acrobatics produce a +15

LebombJames:
This was fixed by #1245. The root cause of the issue was the use of `fromUuidSync` in StackModifiers. When fetching the actor of an unlinked token, this runs data prep, which runs StackModifiers, which runs `fromUuidSync` again. The solution was to just always use the actor passed to StackModifiers instead of fetching it seperately. Thanks for this PR regardless!

PR #1244, opened 2023-08-14
Fixes #1233

### LINKED ISSUE #1233: Sonic vulnerability not taken into account
**Describe the bug**
The vulnerability configured on the actor sheet does not work for Sonic damage. All other damage types seem to work fine.

**To Reproduce**
- create an actor
- under 'Damage Vulnerabilities' select 'Sonic'
- drag this actor in a scene and apply sonic damage with the right click option 'Apply damage' from chat

**Expected behavior**
The Sonic damage should subtract 1.5 times the damage rolled.

**Actual behaviour**
Sonic damage is applied as rolled, without modification.

**Screenshots**
Sonic Vulnerability

 - OS: Win 11
 - Browser: electron app
 - Foundry Version: v11306
 - SFRPG Version: Tested on the development branch, fetched and pulled up-to-date as of today

**Additional context**
Was confirmed by two more people in Discord. Our wild guess is, that the double character 'So' for sonic damage might be the culprit, since all other (single character) damage types work as expected.

Imper1um:
Fixes #1233 

The extra regex ^ was causing the expression to treat the first character as a splitter, but, with how JavaScript Regex Splitters work, it will see that ^ is index 0, and then use that character at index 0 as the split, so if you have a string:

"so"
A regex with ([^,;])+ will create three groups:
"","o",""
This is because it finds that index 0 is a split character, and creates an empty string for the first item, and then puts index 1 as the second item. The third item is something something extra that RegEx likes to do. I'm sure I could look it up, but this just fixes the issue entirely.

### DIFF
--- a/src/module/actor/mixins/actor-damage.js
+++ b/src/module/actor/mixins/actor-damage.js
@@ -89,7 +89,7 @@ export class SFRPGDamage {
     static createDamage(damageAmount, damageTypes = [], isCritical = false, properties = [], options = {}) {
         const parsedDamageTypes = [];
         if (damageTypes.constructor === String) {
-            const splitTypes = damageTypes.trim().split(/([^,;])+/gi);
+            const splitTypes = damageTypes.trim().split(/([,;])+/gi);
             for (const type of splitTypes) {
                 if (type === ',' || type === ';') {
                     continue;


LebombJames:
Included in #1251

PR #1246, opened 2023-08-14
 0.25.1 (#1245)

LebombJames:
* 0.25 hotfixes

* More fixes

* Manifest and changelog


Imper1um:
Does not fix this error:
https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/blob/de533d14e09c791f0966cc4c9f017fdd063eb7ab/src/module/rules/actions/actor/calculate-encumbrance.js#L11
`encumbrance` is called before it is created. It should be `data.encumbrance`

PR #1254, opened 2023-08-17
Solarian Crystal modifiers

Iankid:
Changes Solarian crystals over to applying damage bonuses to parent item.


mullholland:
Is this working?

Iankid:
last I was aware, it was.

LebombJames:
The modifiers look good, but why did you add the critical effects to the descriptions?

PR #1262, opened 2023-08-26
Guard against empty difficulty data

### LINKED ISSUE #1261: [Bug] When rendering combat when difficulty, if difficultyData.difficulty is empty, renderDifficulty crashes
**Describe the bug**
`combat.js` in method `renderDifficulty` will crash if diffObject.difficultyData.difficulty is empty.

This will cause the rendering of combat menus to desync when the difficulty is empty.

**To Reproduce**
1. Create a Combat
2. Try to open the combat.
3. `renderDifficulty` will fail at `difficultyHTML.classList.add` because `difficulty` is empty

**Expected behavior**
Skip `difficultyHTML.classList.add` with variable `difficulty` when `difficulty` is null or empty.

 - OS: Windows 10
 - Browser: Chrome
 - Foundry Version: v11.307
 - SFRPG Version: 0.25.2

Imper1um:
Separates out difficulty add so that if difficulty is empty in any way that it won't attempt to add it as a classList add. This will prevent combat menu desync. Fixes #1261

### DIFF
--- a/src/module/combat/combat.js
+++ b/src/module/combat/combat.js
@@ -629,9 +629,7 @@ export class CombatSFRPG extends Combat {
 
         const difficultyHTML = document.createElement("a");
         difficultyHTML.classList.add("combat-difficulty");
-		if (difficulty) {
-			difficultyHTML.classList.add(difficulty);
-		}
+        if (difficulty) difficultyHTML.classList.add(difficulty);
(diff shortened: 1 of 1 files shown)


LebombJames:
I can't reproduce the initial bug, but this doesn't hurt to have

PR #1274, opened 2023-10-18
EN feats

Iankid:


ian612 (approved):
As before, I reviewed a few of these at random and didn't spot anything erronious, so going to go ahead and merge these. If any errors or omissions are noticed in the future, we'll deal with those on a case-by-case basis.

PR #1288, opened 2024-01-17
Sourcemap issue bugfix

ian612:
This fix addresses an issue where the sourcemaps for the code where getting mangled during build, which would prevent VS Code from hooking in properly during debugging. This would cause issues with breakpoints not binding and prevent use of the external VS Code debugging and inspection tools.

The fix is essentially to rejig how the `*.js` files are pulled together during build, and to set the `sourceRoot` property to the correct source folder when writing sourcemaps.


ian612:
I've had a bit of instability with this, in particular with `build:watch`, so might require some further investigation before it's fully ready

ian612:
I've fixed this with a much better method for getting the sourcemaps to compile correctly. This should allow debugging both directly in Chrome and VS Code without requiring anyone to change their workflow.

ian612:
Closing this because it's not _really_ the best way to implement this, and I have realized that VS Code connects correctly when `build:watch` is being used, as long as at least one change is made and the watch files are copied.

PR #1296, opened 2024-02-08
General Roll Code Cleanup

ian612:
My aim with this is to clean up messy parts of the Roll Code (TM), and document it better with lots of additional comments and explanation. The process of generating a roll is quite the Gordian knot at the moment, and it's difficult for contributors who aren't already familiar with it to make any changes or track down bugs.

I may also add to the actual documentation as part of this, once the cleanup is done.


ian612:
I might still do this at some point, but will leave it closed for now since I didn't make much progress and am not actively working on it

PR #1297, opened 2024-02-11
Fixes for Roll calculations and breakdowns

danimrath:
1) Get ship computer bonus during the initial stage as the context for the starship is not always available when calcs are done.
2) Only add Science officer lock-on option when it is a gunner action
3) pass a key downstream to allow identifying the correct context to use rather than hard coded "actor" which becomes a default.
4) _removeModifierNodes change to only call if the mod has a mod type and allow it to remove nodes if the modifier has no id and the node has a matching name.
5) Add power core and quadrant modifiers to weapon tab and action tab but only when the value isn't 0
6) When looping through the available mods to populate assign a key value of the name of the mod if it does not have an id. This way it does not keep adding the same modifier to nodes if the don't have id's
7) Some simple format changes
8) removed unnecessary escape to regex matching
9) When populating the formula for display just pull out the second to last part of the path to display which should be a name rather than the entire path which is very long.  ex powerCore vs @ship.attributes.systems.powerCore.modOther
10) Do not gather calculated Mods for skill rank variables or ability mod variables those are standalone.
11) When pushing calculatedMods to the rollMods avoid adding a mod that was already added by checking for it's name in rollMods and also in the base formula.


LebombJames:
Given this a test and it all seems to work. My only note is that I don't think we should be doing 9. Showing the full formula, while messy is informative and aids in seeing precisely what is going into a formula, and helps users learn the paths. In the future we could definitely create some "label" system, but as it is I don't want the loss of information.

danimrath:
@LebombJames Give it another go. Should still work and no longer remove the full formula.

PR #1303, opened 2024-03-27
Add a globalThis to make class extension by modules easier

ian612:
Moved the structure defined in `game.sfrpg` to a variable definition `moduleStructure`, then passed that first to `globalThis`, then add `engine` and pass to `game.sfpg` in the init hook


ian612:
I've also confirmed that this doesn't cause any issues with timed effects or the compendium art mapping.

Issue #1192, opened 2023-07-07
Mechs


Imper1um:
In September 2021, Paizo released Tech Revolution (ISBN-13: 978-1-64078-352-2). In this book, it gave the rules for Mechs (starting on Page 90). This system does not fit into the mehcanics of any of the current sheets (NPC, Vehicles, Drones, etc.), as it contains mechanics which are much different than Vehicles, Drones, and NPCs, including Power Core, Power Points, and how it handles armor and auxiliary systems.

Public Rules for Mechs are available on the AONSRD at https://aonsrd.com/Rules.aspx?ID=1973

Iankid:
This is on the list, and the issue post is appreciated. In the mean time you can at least fake mechs in PC/NPC sheets

PR #1307, opened 2024-04-20
Mech Functionality

ian612:
This PR will add mech items and sheets to the system.

Item Sheet Checklist:

- [x] Mech Frame
- [x] Mech Limb
- [x] Mech Power Core
- [x] Mech Power Core Template
- [ ] Mech Weapon
- [ ] Mech Auxiliary System
- [ ] Mech Upgrade

Mech Sheet Checklist

- [ ] Attribute Display
- [ ] Operators
- [ ] Slot display
- [ ] Rolling Attacks & PP Abilities

Logic Checklist:

- [ ] Calculate Attributes
- [ ] Calculate Mech Points
- [ ] Calculate Movement
- [ ] Drag and Drop Item Modifications


ian612:
This will close #1192

ian612:
Some confusing things/errors/omissions/bugs brought up by @rmorgens 

- [x] Can't set a frame to have multiple types
- [ ] Biped, Fast lower limbs has a Sprint PP ability to add 10' enhancement bonus to speed.  Unclear if that should be a situational bonus on the limb, a condition that can be applied since it only lasts 1 round, or if the PP ability should be able to have a modifier
- [x] Power Core Template needs a prerequisite field
- [ ] Unclear how to add the special effect from the Battle core template (temporary boost to PP)
- [ ] Should Upgrades have fields for KAC/EAC/FortSave/ReflexSave/BaseHP/HP Advancement/ShieldPts/Speed?  Or should I just use modifiers for the uppers?  (Modifiers may be challenged if the upgrade is selected multiple times)
- [ ] Weapon Costs are multiple by weapon level, not tier.  Can be level up to Tier+1
- [ ] Will need the weapon properties selections to engage things like Penetration, Reach, Trip, Blast
- [ ] PP ability on Mech Limb doesn't appear to be accepting an equation for the DC.  Trying to set Trample to 12+floor(@details.tier/2).  It will only take a fixed number
- [ ] To Defy the Dragon also introduced frame variants (basically starmetal adjustments to a frame). Only 1 is allowed at a time
- [ ] Mechageddon has a Pyramid frame variant, but I'm instantiating that as a set of Upper Arms, since the variant prevents any upper limbs (among other limitations), so that slot can be used for tracking....  If you want to instantiate more purely, let me know how you approach that problem.
- [ ] Would be useful if an effect could add or subtract PP and Shield Points
- [ ] There are also functions that take ALL the current PP, so some ability to subtract the current would be helpful

apocolips18:
Still cheering on this endeavor! Would love to expand into the mech side for a campaign while still using Foundry.

ian612:
Hi there, sorry still on hold unfortunately while we get Foundry v13 support ready and attend to some other longstanding features and fixups.

AnachronicRodent:
Has there been any further updates on this, I am not very good with understanding Git, but just trying to help my GM find the stuff for Mech for our current game

ian612:
Hi there, currently waiting on a few other things to be finished before resuming work on this. I'll go ahead and make a milestone for it to make it clearer where it stands on the progress track.

PR #1314, opened 2024-05-09
Repaired "picking up dropped items"

Proximitron:
If a dropped item is clicked on and pulled into the inventory, it results in a "processedDroppedData", transfering the instance of the dropped object under "parsedDragData.items[0]". The item is not given as ItemCollection but as an object already, and by casting "toObject()" it simply breaks at this point. As such things can't be taken into the inventory. Removing that wrong command makes it work as intendet.

### DIFF
--- a/src/module/actor/sheet/base.js
+++ b/src/module/actor/sheet/base.js
@@ -1233,7 +1233,7 @@ export class ActorSheetSFRPG extends ActorSheet {
 
         let item = null;
         if (parsedDragData.type !== 'ItemCollection') item = (await Item.fromDropData(parsedDragData)).toObject();
-        else item = parsedDragData.items[0].toObject();
+        else item = parsedDragData.items[0];
 
         // Level up existing class item if dragging on an existing one.
         if (item.type === "class") {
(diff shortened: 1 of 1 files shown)


LebombJames:
I'd rather use optional chaining to only run `toObject` if it is accessible (i.e if the data isn't an object already), because it's important we're mutating a cloned object of the dropped document rather than the original document itself

Proximitron:
I see. If there would be at least a check if "toObject" isn't "undefined", this function would work fine as well. Should i create another merge request or did you do this directly and i can close this?

LebombJames:
I'll just edit this PR and merge it. Thanks!

src/module/actor/sheet/base.js
```diff
@@ -1233,7 +1233,7 @@ export class ActorSheetSFRPG extends ActorSheet {
 
         let item = null;
         if (parsedDragData.type !== 'ItemCollection') item = (await Item.fromDropData(parsedDragData)).toObject();
-        else item = parsedDragData.items[0].toObject();
+        else item = parsedDragData.items[0];
```
LebombJames, line 1236:
```suggestion
        else item = parsedDragData.items[0].toObject?.();
```

Proximitron:
Your change doesn't work, as the optional chain with the "?" leads to an "undefined". As such "item" will be undefined if "toObject" doesn't exist. Doesn't solve the issue.

PR #1315, opened 2024-05-09
second try to fix "dropped item can't be picked up/moved to character"

Proximitron:
This solution checks now if toObject is available, executes it if possible or takes the existing item instance transmitted.

As the purpose of the function is actually to transfer an item, the instance can be changed and/or preserved without issue. If this is still unwanted, a solution would be to use:

item = structuredClone(parsedDragData.items[0]);

### DIFF
--- a/src/module/actor/sheet/base.js
+++ b/src/module/actor/sheet/base.js
@@ -1233,7 +1233,8 @@ export class ActorSheetSFRPG extends ActorSheet {
 
         let item = null;
         if (parsedDragData.type !== 'ItemCollection') item = (await Item.fromDropData(parsedDragData)).toObject();
-        else item = parsedDragData.items[0].toObject?.();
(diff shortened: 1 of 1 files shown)


src/module/actor/sheet/base.js
```diff
@@ -1233,7 +1233,8 @@ export class ActorSheetSFRPG extends ActorSheet {
 
         let item = null;
         if (parsedDragData.type !== 'ItemCollection') item = (await Item.fromDropData(parsedDragData)).toObject();
-        else item = parsedDragData.items[0].toObject?.();
+        else if(parsedDragData.items[0].toObject !== undefined) item = parsedDragData.items[0].toObject();
```
LebombJames, line 1236:
```suggestion
        else if (parsedDragData.items[0].toObject !== undefined) item = parsedDragData.items[0].toObject();
```

PR #1345, opened 2024-08-10
Cook cleanup

ilya-vasiuk:
- changed JournalEntry detection in gulp config
- cleaned up cross-links in docs to avoid cook errors
- renamed item/effects. They have brackets now, so cook does not create new json files and makes changes in proper ones
- recooked all for version 0.27.0


LebombJames:
Gonna close this after our discussion

ilya-vasiuk:
Yes, of course, you are welcome. I would do it on myself, but not sure I can

PR #1357, opened 2024-08-11
DR migration for all creatures

ilya-vasiuk:


ilya-vasiuk:
I see [0.006: Damage Mitigation migration](https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/blob/3fed7677bcf1a51f2e4afd419e28519b0aff128e/src/module/migration.js#L360) should have done it but somehow it didn't

LebombJames:
I suspect that migration was never actually run on the compendium data, and was only implemented as a client-side migration

PR #1361, opened 2024-08-12
Cook cleanup

ilya-vasiuk:


levirak:
This PR looks very similar to #1345, which got closed by @LebombJames "after our discussion." What's the difference?

ilya-vasiuk:
@levirak Yep, we told a bit on Discord. You can see that PR is more broad and contains not only cook fix. It has also changes about JournalEntries, which I migrated to wrong direction, so we decided to close that PR. 
But yes, my main machine is Mac, and I'm still thinking we need to push these changes to make it work for non-windows devs

levirak:
Glad to hear. I mostly asked just to get an explanation on the record.

> we need to push these changes to make it work for non-windows devs

Correct. I ran into that exact problem myself. I've opened #1393 with my own proposed fix. It will conflict with yours once either is merged, however.

ian612:
The changes made to the build script don't seem to result in any issues with building from the Windows side as far as I can tell from my testing.

@levirak I unfortunately don't have a linux or Mac environment to test on, but can you confirm if this branch fixes the build issues on linux as well? I would assume so given that both are Unix-based, but best to test just in case. If it does fix them, then we can close #1393 once this is merged.

levirak:
`npm run build` works\*, but `npm run unpack` fails. \*But I still have my issues with the implementation.

For build, both the original and this build file do string substitutions to manipulate paths. The original was looking for `\\src`, and this one looks for just `src`. That works, as it no longer looks for an OS specific path separator, but it's now even less protected string manipulation of a path. Instead I'd prefer to use Gulp's built-in way to achieve what this is trying to do: 
```javascript
gulp.src(sources, { base: 'src' })
    .pipe(gulp.dest('dist'));
```

For unpack, the issue is that Linux (but not MacOS) file systems are usually case sensitive. To make it work on my machine I made this change:
```diff
 async function unpackPacks(partOfCook = false) {
-    const sourceDir = partOfCook ? "./src/packs" : `${getConfig().dataPath.replaceAll("\\", "/")}/data/systems/sfrpg/packs`;
+    const sourceDir = partOfCook ? "./src/packs" : path.join(getConfig().dataPath, 'Data/systems/sfrpg/packs');
     console.log(`Unpacking ${partOfCook ? "" : "and sanitizing "}all packs from ${sourceDir}`);
```

#1393 addresses both of these concerns. It conflicts with changes here, but they would be trivial to resolve. Or to just do in this PR, if people would prefer.

danimrath:
I don't have Linux or Mac but it is working fine for me.

danimrath (approved):
This is working fine on my computer though it is Windows OS.

ian612:
@levirak Since this PR has some additional stuff that's needed for the `cook` data entry process to get fixed up, I think what I'll do is merge this, then merge #1393 right afterward, overwriting the conflicts with your suggested changes, as they seem to work fine together on my Windows build. Thanks for the suggestions!

Issue #1212, opened 2023-07-22
FR: Add DC functionality to `@Check`


avagdumortesin:
**Is your feature request related to a problem? Please describe.**
When inputting content as the GM for an adventure, I like to add `@Check` calls for skill checks and saving throws, because I enjoy the automation abilities. 

**Describe the solution you'd like**
It would be awesome to add the ability to add a DC to the `@Check` so that a quick glance at the roll card provides the success/failure information.

**Additional context**
This would, in my imagination, work similarly to the PF2e system `@Check` call, so entering something like `@Check[type:computers|dc:17]{Computers check}` would show a Success or Failure in the roll card, depending on the results of the d20 roll and the actor's Computers bonus.

ian612:
Closed by #1369

PR #1369, opened 2024-08-15
Add optional DC parameter to checks

ilya-vasiuk:
My thoughts about #1212 

You can see `@Check[type:strength|dc:10]` and `@Check[type:medicine|dc:10|displayDC:true]` on the screenshot
Screenshot 2024-08-16 at 16 20 53
Screenshot 2024-08-16 at 16 43 32


ian612:
This looks good to me!

PR #1375, opened 2024-08-17
Effects regeneration

ilya-vasiuk:
Recooked files and renamed them properly, cause Foundry didn't pick them up and there was only one effect available


ian612:
These changes got merged into development via another PR somewhere (I think it might have been on your re-cook everything PR), so going to close this one.

PR #1319, opened 2024-06-24
Starfinder Enhanced Themes

MrPaolinsky:
- [ ] Personal Trainer
- [X] Bouncer
- [X] Paranormal Investigator (Enhanced)
- [X] Pickpocket
- [X] Rancher
- [X] Starwalker (Enhanced)
- [X] Street Magician 
- [X] Vidgamer

PR #1383, opened 2024-08-21
allow blank theme skill and attribute to be selected

### LINKED ISSUE #1382: Themes should provide a blank option for skill selection and ability adjustment
Some new themes, like Personal Trainer, don't provide a skill bonus and instead provide an alternate feature. Some may also offer a choice between abilities.

There should be an option to select a blank skill and/or ability when creating the theme.

ian612:
Resolves #1382, and will allow #1377 and/or #1319 to have data input correctly

### DIFF
--- a/src/template.json
+++ b/src/template.json
@@ -2104,7 +2104,7 @@
         "modifiers"
       ],
       "abilityMod": {
-        "ability": "cha",
+        "ability": "",
         "mod": 1
       },
       "skill": ""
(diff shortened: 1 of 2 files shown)


src/template.json
```diff
@@ -2104,7 +2104,7 @@
         "modifiers"
       ],
       "abilityMod": {
-        "ability": "cha",
+        "ability": "",
         "mod": 1
```
danimrath, line 2108: Maybe blank or 0 for mod? I don't THINK it's causing any issues but feels odd to have a default mod if no default ability.

danimrath (approved):
Left a suggestion but otherwise don't see any issues.

src/template.json
ian612: I think for almost all (if not all) themes, the default is a bonus of +1, so I want to leave that as the default.

PR #1389, opened 2024-09-11
Roll Deprecated (Fix for IsRoll and DiceSoNice integration)

danimrath:
roll has been deprecated in favor of rolls when doing ChatMessage.Create. When rolls:[] is not used the chat message will not flag itself as a roll when IsRoll is called which can break all kinds of functionality like DiceSoNice knowing there is a dice roll to render.

### DIFF
--- a/src/module/actor/actor.js
+++ b/src/module/actor/actor.js
@@ -919,7 +919,7 @@ export class ActorSFRPG extends Mix(Actor).with(ActorConditionsMixin, ActorCrewM
             speaker: ChatMessage.getSpeaker({ actor: speakerActor }),
             content: rollContent,
             rollMode: rollMode,
-            roll: rollResult.roll,
+            rolls: [rollResult.roll],
             type: CONST.CHAT_MESSAGE_STYLES.ROLL,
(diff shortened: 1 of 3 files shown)


ian612:
I tested it a bit and it seems to respect the roll visibility setting in the sidebar, but not the one selected in the popup dialog box. There might be something in the dialog application that needs to be tweaked as well?

image

ian612 (changes requested):
Fixes the sidebar setting for rolls, but dialog box roll visibility setting still seems to be ignored

danimrath:
> Fixes the sidebar setting for rolls, but dialog box roll visibility setting still seems to be ignored

Interesting. I wasn't actually trying to fix that specifically. I was fixing dice so nice and things trying to use the IsRoll flag. If you think that is related I can take a look later today though.

ian612:
Ah woops, yep you're right the issue is different, so we should do a separate fix for what I was noticing. I'll test this out again later today and just verify the Dice So Nice functionality

ian612 (approved):
After testing, I see no issues and I think this is good to merge whenever you're ready to!

PR #1400, opened 2024-11-11
Add drag-and-drop macro creation for character stats

levirak:
Allows for dragging an actor's skill, ability or save into the macro bar to create a macro to roll the check.

This was a feature requested by a player in a Starfinder game I'm in, so I figured I'd offer this change to the project.

Marked as a draft PR because I'm unsure where the hook code should live, because I'd like advice on if I should write new globally registered functions (like `game.sfrpg.rollItemMacro(itemId, type)`) and because I'm fishing for some better icon suggestions.


src/module/actor/sheet/base.js
```diff
@@ -1350,3 +1389,93 @@ export class ActorSheetSFRPG extends ActorSheet {
...
+    str: 'icons/magic/control/buff-strength-muscle-damage-orange.webp',
+    dex: 'icons/skills/movement/feet-winged-boots-brown.webp',
+    con: 'icons/svg/d20.svg',
+    int: 'icons/svg/d20.svg',
+    wis: 'icons/svg/d20.svg',
+    cha: 'icons/svg/d20.svg',
+};
+
+// TODO(levirak): where should this live?
+Hooks.on("hotbarDrop", async (bar, data, slot) => {
```
LebombJames, line 1435: This should probably live in the `item-macros.js` file, and that file should be renamed something more general, maybe just `macros` or `hotbar-macros` or something

levirak: Thanks for the suggestion!

levirak:
I've moved the macro creation code into `item-macros.js` and renamed that file to `hotbar-macros.js`. Git, regrettably, has recorded this as a delete-create rather than a rename-edit. Oh well.

I've also done my best to pick decent icons for each check. I'm mostly happy with them, though I wish there were better ones available. Suggestions are welcome.

ian612:
If you need suggestions for icons, I think the inline text enricher (`src/module/system/enrichers/check.js`) has a bunch of font-awesome icons defined for skills. You could use those for consistency with journal checks if you want

levirak:
That's a very good suggestion. Though, is there a way to use text for the hotbar macro image? I'm under the impression that that can only be an image. I could plausibly save each FA glyph as an SVG image and use those, though I'm uncertain if that would have copyright implications.

levirak:
Well, scrap what I was saying about making SVG files and copyright. It looks like Font Awesome has options to let you download the glyphs as individual SVG files. I still believe they would need to live as files, rather than just being passed as strings. Is that correct? And is it worth while to add these glyphs to the repo, probably as `src/icons/fa-svg/*.svg`?

levirak:
For reference, here's a comparison between using the random icons I chose and what it looks like to use FA icons.

image image

One thing I'll note is that it doesn't seem like I can just use FA's provided SVG files because they aren't all square. Some pre-processing will be necessary to make the icons square, rendered in white, and perhaps to add some padding or background fill.

Again, unless there's a way to use the font directly.

ian612:
Cool, looks very nice. My slight preference is for the fa icons, since the others are used for various feats and features and such within the compendiums, so that differentiates them a bit from each other visually, but as it's your feature I'll defer to whatever you think looks better!

levirak:
I think I prefer the FA icons too. (Especially now that I've put in the work to automate making them.) I only hesitate because my current implementation would add one new script, two new development dependencies and a new directory full of images.

I've only just noticed that `src/templates/ui/hotbar.hbs` defines the hotbar, so before I commit to that, I think I'll look into what it would take to just render text in the hotbar instead of an image.

levirak:
Well. Final report time, I suppose.

|Method|Example|
|:--|:--|
| Existing Image | image |
| Square FA SVG | image |
| Uniform Size FA SVG | image |
| Text | image |

I've decided to commit to the uniformly sized SVG approach. It has all the same advantages and disadvantages of the non-uniform approach, except it also keeps some icons (such as Intimidate in slot 9) from feeling too big. My only issue is that some narrow icons (such as Acrobatics in slot 1) feel too small. The result is nearly identical to using the font directly.

In many ways I'd consider using text to be ideal, but it's not worth the implementation, especially for such a one-off feature. Getting a FA glyph scaled and centered in its slot is finicky, and getting the hotbar to handle both images and FA class names convolutes everything a bit too much for my tastes. And unless I'm prepared to extend `Macro`, even attaching a FA class name to a macro requires using flags in a way that I don't appreciate.

In all, rather than eat the runtime complications of handling both FA glyphs and images, I'd rather eat the development complications of pre-processing FA into correctly scaled and centered SVG images.

I will miss the mosquito in amber icon for Life Science, however.

levirak:
Is there anything holding this PR up?

ian612:
Hi there, nope nothing holding it up afaik, just need to do a final review and merge. I'll try to take care of that this weekend or next!

src/module/system/hotbar-macros.js
```diff
@@ -0,0 +1,284 @@
...
+ * @returns {Promise<Macro>}
+ */
+function findElseCreateMacro(data) {
+    const existingMacro = game.macros.contents
+        .find(m => (m.name === data.name) && (m.command === data.command));
+
+    return existingMacro
+        ? Promise.resolve(existingMacro)
+        : Macro.create(data, { displaySheet: false });
+}
```
LebombJames, line 34: Think we can just make this an async function and return `existingMacro` or call `Macro.create`, without manually creating promises

levirak: Yep. On second look, this can be greatly simplified.

ian612:
Functionally, this seems to work correctly to me with the characters I've tested it with. @LebombJames if you don't have any other issues with the code that require changing, I'd like to merge this.

Issue #1291, opened 2024-02-01
V12 Todo list


LebombJames:
- [x] Replace all global scope foundry util function calls with properly namespaced calls. (e.g `duplicate` -> `foundry.utils.duplicate`
- [x] Defer to new core measurement tools
- [x] Replace references of roll term classes to properly namespaced calls
- [x] Unfuck synchronous roll evaluations (async rolls 🥳)
  - [x] Partially: `DiceSFRPG.simplifyRollFormula` returns `NaN` for paranthetical terms.
- [ ] Migrate from `select` to `selectOptions` in Handlebars files.
- [x] Change path for global canvas lumionsity in `vision.js`
    - Still to do: update vision modes with new features for Darkness, luminosity etc.
- [x] Change path for default actor model in `tokendocument.js`
- [x] Update `_preUpdate`s to account for the whole document being present, not just the diff
    - Seemed to work as is? Strange
- [x] Migrate or patch roll parser to account for unbracketed functions preceding dice (`floor()d6`)
- [x] Condition/effect Icons are not appearing on tokens
- [x] Token Condition Hud is missing condition names

BroderPeters:
Are the two remaining ones already worked on and can I support or try to implement them myself? (haven't worked on the project before)

ian612:
`selectOptions` was added in #1401, so I'm going to close this out as completed.

PR #1401, opened 2024-11-14
migrate `{{#select}}` to `{{selectOptions}}`

levirak:
I got sick of seeing the warnings in my development console, so I decided to do something about it. This should also close out the last item of #1291's checklist.

I chose to create a `{{selectOption}}` helper that would add ad hoc options with a similar API to `{{selectOptions}}`. `{{selectOption}}` could be avoided by hoisting the extra options into `config.js`, but that implies more sweeping changes than I wanted to commit to. The result is somewhat janky, and probably contradicts Foundry's recommendations, but it greatly eased the transition away from `{{#select}}` and permitted relatively minimal code changes.

### DIFF
--- a/src/module/config.js
+++ b/src/module/config.js
@@ -426,6 +426,13 @@ SFRPG.skills = {
     "sur": "SFRPG.SkillSur"
 };
 
+SFRPG.controlSkills = {
+    "pil": "SFRPG.SkillPil",
+    "ath": "SFRPG.SkillAth",
+    "sur": "SFRPG.SkillSur",
+    "none": "SFRPG.None"
+};
+
 // Weapon Types
 SFRPG.weaponTypes = {
     "basicM": "SFRPG.WeaponTypesBasicMelee",
@@ -1958,6 +1965,12 @@ SFRPG.droneBadSaveBonusPerLevel = [0, 0, 1, 1, 1, 1, 2, 2, 2, 2, 3, 3, 3, 3, 4, 
 SFRPG.droneFeatsPerLevel = [1, 1, 2, 2, 2, 3, 3, 3, 4, 4, 5, 5, 5, 6, 6, 6, 7, 7, 8, 8];
 SFRPG.droneModsPerLevel = [1, 1, 2, 2, 3, 3, 4, 4, 5, 5, 6, 6, 7, 7, 8, 8, 9, 9, 10, 10];
 SFRPG.droneAbilityScoreIncreaseLevels = [4, 7, 10, 13, 16, 19];
+
+SFRPG.droneArmTypes = {
+    "general": "SFRPG.DroneSheet.Mod.Details.Arms.ArmType.General",
+    "melee": "SFRPG.DroneSheet.Mod.Details.Arms.ArmType.Melee",
+    "ranged": "SFRPG.DroneSheet.Mod.Details.Arms.ArmType.Ranged"
+};
 
 SFRPG.capacityUsagePer = {
     "action": "SFRPG.Capacity.UsagePer.Action",
--- a/src/templates/apps/movement-config.hbs
+++ b/src/templates/apps/movement-config.hbs
@@ -23,11 +23,7 @@
         <div class="form-fields">
             <input name="system.attributes.speed.flying.base" type="number" step="0.1" value="{{system.attributes.speed.flying.base}}"/>
             <select name="system.attributes.speed.flying.baseManeuverability">
-                {{#select system.attributes.speed.flying.baseManeuverability}}
-                <option value="-1">{{ localize "SFRPG.ActorSheet.Attributes.Speed.Flight.Clumsy" }}</option>
-                <option value="0">{{ localize "SFRPG.ActorSheet.Attributes.Speed.Flight.Average" }}</option>
-                <option value="1">{{ localize "SFRPG.ActorSheet.Attributes.Speed.Flight.Perfect" }}</option>
-                {{/select}}
+                {{selectOptions (config "flightManeuverability") selected=system.attributes.speed.flying.baseManeuverability}}
             </select>
         </div>
     </div>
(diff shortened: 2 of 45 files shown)


src/module/config.js
```diff
@@ -426,6 +426,13 @@ SFRPG.skills = {
     "sur": "SFRPG.SkillSur"
 };
 
+SFRPG.controlSkills = {
+    "pil": "SFRPG.SkillPil",
+    "ath": "SFRPG.SkillAth",
+    "sur": "SFRPG.SkillSur",
+    "none": "SFRPG.None",
```
danimrath, line 433: trailing comma

src/module/config.js
```diff
@@ -1959,6 +1966,12 @@ SFRPG.droneFeatsPerLevel = [1, 1, 2, 2, 2, 3, 3, 3, 4, 4, 5, 5, 5, 6, 6, 6, 7, 7
 SFRPG.droneModsPerLevel = [1, 1, 2, 2, 3, 3, 4, 4, 5, 5, 6, 6, 7, 7, 8, 8, 9, 9, 10, 10];
 SFRPG.droneAbilityScoreIncreaseLevels = [4, 7, 10, 13, 16, 19];
 
+SFRPG.droneArmTypes = {
+    "general": "SFRPG.DroneSheet.Mod.Details.Arms.ArmType.General",
+    "melee": "SFRPG.DroneSheet.Mod.Details.Arms.ArmType.Melee",
+    "ranged": "SFRPG.DroneSheet.Mod.Details.Arms.ArmType.Ranged",
```
danimrath, line 1972: trailing comma

src/module/config.js
levirak: I like trailing commas. Don't you? They make latter additions and deletions easier and keeps diffs / git-blames more meaningful. When I've got multi-line objects like this, I think of commas like semicolons.

Though I've only now noticed the `.eslintrc.json` file, which is close enough to a style guide and considers it an error, so I'll change it.

By the way, eslint on develop reports 273 errors and 177 warnings.

danimrath:
I only mentioned the comma because of the eslint error. I've only just started looking at PR's and the broader project so I don't have a particular opinion on that one :). I didn't have time to finish reviewing all the changes in the PR but I was able to quickly load things up and nothing seemed off.

src/module/config.js
ian612: Are these additions directly related to the replacement of select with selectOptions?

ian612: Ah wait I see, they're adding in values that used to be hardcoded into the sheets 👍

src/templates/apps/movement-config.hbs
```diff
@@ -47,14 +43,7 @@
...
             <select name="system.attributes.speed.mainMovement">
-                {{#select system.attributes.speed.mainMovement}}
-                <option value="land">{{ localize "SFRPG.ActorSheet.Attributes.Speed.Types.Land" }}</option>
-                <option value="burrowing">{{ localize "SFRPG.ActorSheet.Attributes.Speed.Types.Burrowing" }}</option>
-                <option value="climbing">{{ localize "SFRPG.ActorSheet.Attributes.Speed.Types.Climbing" }}</option>
-                <option value="flying">{{ localize "SFRPG.ActorSheet.Attributes.Speed.Types.Flying" }}</option>
-                <option value="swimming">{{ localize "SFRPG.ActorSheet.Attributes.Speed.Types.Swimming" }}</option>
-                <option value="special">{{ localize "SFRPG.ActorSheet.Attributes.Speed.Types.Special" }}</option>
-                {{/select}}
+                {{selectOptions (config "speeds") selected=system.attributes.speed.mainMovement}}
```
ian612, line 46: Here you have `(config "speeds")` and in `chassis.hbs` you have `config.speeds` replacing the same list of options, is there a difference between them?

levirak: The difference is that the `config` variable wasn't available in this template because `src/module/apps/movement-config.js` didn't return it from its `getData()` method. I could easily just add it, but when I was doing this work I was concerned I'd run into exactly this problem several times. Rather than hunt down every `getData()` method that needed to be updated, I decided to register the `config` helper (a better name may be warranted) that would guarantee access to `CONFIG.SFRPG`. At the same time, I didn't bother to replace existing instances of the `config` variable, and I went back and forth a bit on which I preferred.

Configuration is such a static, cross cutting concern that I think the helper should be preferred over passing it as data, but I'm not really adamant about that stance.

danimrath (approved):

danimrath (approved):

src/templates/apps/movement-config.hbs
ian612: Ah gotcha, I didn't realize it was a new helper. While I actually like the idea of using a config handlebars helper for that data, I think for this PR we should stick with the `getData()` method of accessing it. This is primarily for the sake of keeping the codebase consistent; unless there's a strong case for using a different method for an individual sheet, it's preferable to use one method across the whole project rather than having two different ways of accessing that data.

If it makes more sense to switch all of the sheets over to using a `config` helper, then I'm definitely not opposed to doing that, we should just transition all the sheets at the same time if we do switch, probably in a separate PR to keep the scope of changes as narrow as it can be. We should also consider, when updating sheet functionality, that Foundry core is transitioning toward their AppV2 framework and phasing out AppV1 slowly. While I'm not that knowledgeable on V2 specifics right now since the deprecation period is long, we should probably start evaluating all changes we make to sheets against the requirements for that framework.

ian612:
@levirak I believe danimrath has made the changes I mentioned [here](https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/pull/1401#discussion_r1898959418), to switch passing the `CONFIG` data to sheets over to using the `getData()` method. If you're happy with how the PR looks, I'm satisfied with how this is set up.

Thanks again for submitting this! My VS Code debug console is in your debt lol

levirak:
It looks like all the uses of the helper have been removed, so its declaration in `src/module/handlebars.js` should also be removed. It's a very simple helper anyway, so if there's an interest in switching all the variables to helpers, it can be recreated that PR.

danimrath:
> It looks like all the uses of the helper have been removed, so its declaration in `src/module/handlebars.js` should also be removed. It's a very simple helper anyway, so if there's an interest in switching all the variables to helpers, it can be recreated that PR.

It makes sense so I've removed it until the following PR.

PR #1402, opened 2024-12-13
Fix bug preventing combat type switch

ian612:
Fixes an issue in 0.27 where clicking the combat type switch button would erroneously call the routine to handle timed effects and throw an error.


danimrath:
crashes if undefined.
        if (!changed.flags?.sfrpg?.combatType && options.eventData) {
            this._handleTimedEffects(options.eventData);
        }

ian612:
Thanks for the catch, fixed

PR #1403, opened 2024-12-16
Starship Combat Initiative Fixes

danimrath:
This fix allows having no pilot and also applies piloting bonus when rolling initiative
1) Set initiative to piloting bonus if no pilot
2) Include piloting bonus when rolling for initiative.


danimrath:
Just realized it needs to still apply bonuses.

ian612 (commented):
Looks good to me! I think it's fine to merge

PR #1404, opened 2024-12-19
Update data source codes and formatting for contributors

ian612:
Updated the list of sources on the contributing file to include the most recent set of hardcover releases, and to reorder books so they are in chronological release order.


danimrath (approved):
Seems fine to me.

PR #1407, opened 2024-12-19
Run unpack to get recent data changes

ian612:
When #1361 was merged, it looks like `unpack` hadn't been run on the full set of data `.json` files in a while, and I forgot to run it before committing. This meant a few data structure changes that had been implemented since that PR was authored are now causing `unpack` to show every file as changed whenever someone makes data updates.

This is just a simple `unpack`, but should be a final fix for this problem that's preventing data updates from being made smoothly.


danimrath (approved):

danimrath (commented):
Cooking results in errors.

ian612:
Interesting, it doesn't seem to result in any issues when I run `cook` on it. What error is it throwing, and are you cooking while the Foundry world is open or closed?

danimrath (approved):

danimrath:
> Interesting, it doesn't seem to result in any issues when I run `cook` on it. What error is it throwing, and are you cooking while the Foundry world is open or closed?

Weirdest thing. I was trying all kinds of things and it just kept giving an iteration error. I definitely did not have foundry open. I managed to fix it by running VS code as admin then doing npm run clean, a build and then a cook.

Issue #1188, opened 2023-06-30
Common Effect Items


Iankid:
This issue post is to collect candidates for the upcoming Effect Item feature.
There is a long list of features/abilities/items that can make use of it. So for now I want to gather a list of known ones.
Commonly used ones will be a preference, but not limited to.
The effect item will act similarly to conditions, adding modifiers to the host actor. The bonus of Effect items is that a Duration and even Origin actor can be listed.

Please reply with your suggested Effect Items.

![image](https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/assets/60759323/22494b2a-b58b-49c7-8473-ebe8951f94fd)

Iankid:
-Biohacker's Biohack feature
-Envoy's Get em' feature

ian612:
Precog
- Chronomatic defense
- Chronomatic flow

Mordaedil:
Archetype Starsinger

- Gift of Song

rmorgens:
Envoy Features
  - Brace Yourself

sirbutter:
I will take this task. For now I will search every ability that needs effects, starting from feats. My criteria is any ability that has any temporary property that is transfered to other character or can't be solved as situational modifier to checks (such as AC changes). I will also add some other effects to track targets of some other abilities, like Get 'em target.
Feel free to add suggestions.

- Cleave
- Lunge
- Mobility
- Rallying Roar
- Slippery Shooter
- Combat Casting
- Epic Retort
- Borrowed Vitality
- Bodyguard
- Craven Ploy
- Crescendo of Violence
- Combat Casting

PR #1408, opened 2024-12-20
Starfinder Enhanced equipment

rmorgens:
Adds Equipment from Starfinder Enhanced - Augmentations, Magic Items, Technological Items, Hybrid Items, Grenades, Serums, Personal Items


ian612:
Some items look like they could use an effect item to apply their effects in an automated way. No need to do it now, but at some point we should make effect items for these. I'll try to add these to the effect item tracking post as I spot them (Issue #1188)

ian612 (changes requested):
First things first, thanks very much for taking the time to put all of these in! I've gone through everything alphabetically up through the items starting with C and left comments on everything I noticed. In general, most stuff looks good but there are a decent number of items needing activation details added. I also noticed a few instances where setting limited uses would make more sense than setting up ammunition settings.

If you can take a second pass through the items to check for these issues, I'll review the rest of the items afterward and we can get everything merged in!

ian612 (approved):

ian612:
Initially I was trying to look through everything in detail on these large data PRs, but it's become clear to me that that likely isn't realistic from a time-commitment standpoint. I think the best course of action for these from now on is I'm going to spot-review a few items in the PR, make general comments on those, allow for changes to be made, do a second round of spot checks, and then merge if everything looks good. Then we'll rely on community feedback to make any fixes that need to be made.

For this PR in particular, it looks like all changes I requested were made, so I'm going to go ahead and merge. Again I apologize profusely for how long this took to get merged, hopefully expediting the reviews for data PRs will fix that going forward.

Issue #1385, opened 2024-08-22
Rolls from character sheet don't respect Roll Mode when "Use custom Starfinder chat Card" is enabled


TheGreatRedDuke:
**Describe the bug**
When the Use custom Starfinder chat Card? option is enabled, seemingly any roll coming from a character sheet ignores the Roll Mode that is currently set.

**To Reproduce**
1. Create a new world.
2. Check the "Use custom Starfinder chat card" option under the Starfinder config
3. Create a generic Player Character actor
4. Change the Roll Mode dropdown above the chat to "Blind GM Roll"
5. Click any skill on this new character's sheet and change the roll mode to "Blind GM Roll"
6. Roll
7. See the Roll in chat with the wrong Roll Mode

**Expected behavior**
The roll should appear in chat with the correct Roll Mode.

**Screenshots**

**Technical details (please complete the following information):**
 - OS: Windows 11
 - Browser: chrome, electron app (both latest)
 - Foundry Version: 12.331
 - SFRPG Version: 0.26.1

**Additional context**
I'm also noticing that when you click a skill with "Use custom Starfinder chat card" disabled, changing the Roll Mode right above the chat does change the roll mode as expected, but it always seems to overwrite whatever option is selected in the Roll modal, making that selection never work.

Turning off "Use custom Starfinder chat card" in my games is a temporary solution to sidestep this issue, but I _would_ like to use it and have any rolls from a character sheet and their roll mode be respected.

Additionally, though it might not be related, when "Use custom Starfinder chat card" is enabled and Roll Mode is not being respected, Dice So Nice dice rolls are not appearing on screen either if that is enabled.

ian612:
Fixed by #1410

Issue #502, opened 2022-01-12
BUG: Macro bar attacks are always public


CptTwinkie:
Attacks dragged to the macro bar
![image](https://user-images.githubusercontent.com/2079751/149209490-9181cda1-2808-4cbb-9597-cdd2a6b44093.png)
They are always public, ignoring your default setting above the message box

This is tragic for GMs who don't want to share their weapon's stats to keep a little bit of mystery alive in the world

danimrath:
Fixed by [#1410](https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/pull/1410)

PR #1410, opened 2024-12-21
fix roll mode not set properly from dialog box

ian612:
Small tweak to fix roll visibility settings not being set from the dialog boxes correctly

### DIFF
--- a/src/module/item/mixins/item-activation.js
+++ b/src/module/item/mixins/item-activation.js
@@ -60,7 +60,6 @@ export const ItemActivationMixin = (superclass) => class extends superclass {
         updateData['system.isActive'] = active;
 
         const updatePromise = this.update(updateData);
-        const rollMode = game.settings.get("core", "rollMode");
 
         if (active || this.system.duration.value || this.system.uses.max > 0) {
             updatePromise.then(() => {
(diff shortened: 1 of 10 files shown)


ian612:
Should resolve #1385

src/module/item/mixins/item-activation.js
```diff
@@ -60,7 +60,6 @@ export const ItemActivationMixin = (superclass) => class extends superclass {
         updateData['system.isActive'] = active;
 
         const updatePromise = this.update(updateData);
-        const rollMode = game.settings.get("core", "rollMode");
```
ian612, line 63: I think it might be best to keep this to allow for sending of text as private to the GM, if needed

src/module/item/mixins/item-activation.js
```diff
@@ -108,17 +106,6 @@ export const ItemActivationMixin = (superclass) => class extends superclass {
...
-                        chatData["whisper"] = ChatMessage.getWhisperRecipients("GM");
-                    }
-                    if (rollMode === "blindroll") {
-                        chatData["blind"] = true;
-                    }
-                    if (rollMode === "selfroll") {
-                        chatData["whisper"] = ChatMessage.getWhisperRecipients(game.user.name);
-                    }
-
                     ChatMessage.create(chatData, { displaySheet: false });
```
ian612, line 109: After some digging, it looks like this whole section can be replaced with if we want to keep the private text functionality
```js
ChatMessage.applyRollMode(chatData, rollMode);`
ChatMessage.create(chatData, { displaySheet: false, rollMode: rollMode });
```

danimrath:
Should resolve [#502](https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/issues/502)

PR #1416, opened 2024-12-23
Default icons

ian612:
This adds default icons for each actor and item type that get applied automatically in `_preCreate()`.

Once this is merged, I will make an additional PR to alter all the compendium items that still use the default cowled man/bag icon, but only after the outstanding data entry PRs have been completed to prevent conflicts with any that have icons selected.


ian612:
Here are the icons I've selected, for easy review.

## Items
image
image
image

## Actor
image

rmorgens:
Your Vehicle and Vehicle Attack look better for Mechs when you finish that.  The bus under delapouite looks like a better generic vehicle.  The great-war-tank under cathelineau could be vehicle attack.  Let me know if you need me to forward the specific file and indicate PNG or SVG as preferred

ian612:
Ah yes good point on the mech stuff. Appreciate the suggestion, I'll tweak those ones 🤙

danimrath (approved):
Looks good to me!

PR #1415, opened 2024-12-23
Starship Combat Updates

danimrath:
This is to bring the old 650-651 together into one updated branch.
https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/pull/650
https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/pull/651

Adds option to auto roll critical damage and notifications about both critical hits and critical damage.
Cleanup Roll Formula regex to properly remove + + or + - from the final formula.
eslint fixes for ablative
Remove piloting Bonus for NPC crew actions as these are already baked in.
Only add critical effect modifier to formula if non-zero
Add system setting for auto roll starship critical damage effect
Add system setting for hiding hostile starship critical messages

image

PR #1417, opened 2024-12-23
Starship Combat Cleanup round two

danimrath:
Cleanup Roll Formula regex
eslint fixes for ablative
Remove piloting Bonus for NPC crew
Only add critical effect modifier to formula if non-zero


danimrath:
Rolling this into https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/pull/1415.

PR #1422, opened 2024-12-28
Fix default icons always applied on import from compendium

ian612:
The original implementation of the default icons feature had a bug which caused a default icon to always be applied to an actor or item imported from a compendium, even if it had it's own unique icon selected. This update changes the behavior to first check if the actor/item's icon is the default mystery man or item bag, and only replace the icon if that is the case.


danimrath (approved):
Working as intended.

PR #1424, opened 2024-12-29
Compendium icons update

ian612:
This follows #1416 and migrates all existing compendium items and actors (and their prototype token images and nested item images) to use the new set of default icons that's been added to the system.


danimrath (approved):
This seems correct now.

PR #1427, opened 2024-12-31
Define modifiers as SFRPGModifier

danimrath:
When not contained on a parent actor modifiers were throwing errors when attempting to toggle the modifier state using the checkbox.

With this change the toggle functionality should work when on a standalone item or in the compendium.


ian612 (commented):
Looks good, no issues noticed on my end after testing

ian612 (approved):

PR #1430, opened 2025-01-01
Also apply allModifiers against static dc's.

danimrath:
When spells that specify a spell DC are on actors that have modifiers the modifiers do not get applied.

Ex. An NPC with spells with dc's of 15 if you apply a modifier from for example Ego Whip to that NPC with a constant modifier of -2 the DC's of the spells stay at 15. 

This change instead adds allModifiers with a type of spell-save-dc to spell dc's regardless of if the DC is specified or not.


ian612 (approved):
Looks good to me, I see no issues on testing.

Optionally, I think it would probably be worth adding a few additional comments to this file to explain in plain text how the ability prioritization & calculation is done.

Issue #692, opened 2022-07-02
Armor turn penalties not being applied to starships 


Leonaquilla:
The "turn distance penalty" value in the armor starship components is not currently being applied to the turn distance value on the starship's main sheet display.

danimrath:
Fixed in [#1433](https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/pull/1433)

PR #1433, opened 2025-01-02
Include armor turn modifier

danimrath:
When calculating the turn value of a starship include modifiers from armor and ablative armor penalties.

Fixes [#692 ](https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/issues/692)

### DIFF
--- a/src/module/rules/actions/actor/starship/calculate-starship-frame.js
+++ b/src/module/rules/actions/actor/starship/calculate-starship-frame.js
@@ -157,10 +157,11 @@ export default function(engine) {
             const frame = frames[0];
 
             data.frame = frame;
+            const maneuverability = frame.system.maneuverability;
 
             data.details.frame = frame.name;
             data.details.size = frame.system.size;
-            data.attributes.maneuverability = frame.system.maneuverability;
+            data.attributes.maneuverability = maneuverability;
             data.attributes.damageThreshold = {
                 value: frame.system.damageThreshold.base,
                 tooltip: []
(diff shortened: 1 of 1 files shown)


src/module/rules/actions/actor/starship/calculate-starship-frame.js
```diff
@@ -180,8 +181,24 @@ export default function(engine) {
...
+                data.attributes.turn.tooltip.push(`${armorItems[0].name}: ${armorItems[0]?.system?.turnDistancePenalty.signedString()}`);
+                armorTurnPenalty += armorItems[0]?.system?.turnDistancePenalty;
+            }
+
+            if (ablativeArmorItems?.length > 0) {
+                data.attributes.turn.tooltip.push(`${ablativeArmorItems[0].name}: ${ablativeArmorItems[0]?.system?.turnDistancePenalty.signedString()}`);
+                armorTurnPenalty += ablativeArmorItems[0]?.system?.turnDistancePenalty;
+            }
+
+            data.attributes.turn.value = turnManeuverability + armorTurnPenalty;
```
ian612, line 200: This value should be clamped to reflect the maneuverability scale only going from 0 (Perfect) to 4 (Clumsy)

src/module/rules/actions/actor/starship/calculate-starship-frame.js
```diff
@@ -180,8 +181,24 @@ export default function(engine) {
...
+            data.attributes.turn.tooltip.push(`${data.details.frame}: ${turnManeuverability.signedString()}`);
```
ian612, line 201: I think I prefer if items that don't influence the value to not show up in the tooltips for clarity (i.e. if their bonus is +0 they don't appear), but open to leaving it if you think it's better to have that information appear.

src/module/rules/actions/actor/starship/calculate-starship-frame.js
```diff
@@ -180,8 +181,24 @@ export default function(engine) {
...
         }
 
         /** Ensure pilotingBonus exists. */
```
ian612, line 204: I think the ship's turn value should influence its Maneuverability (`@attributes.maneuverability`), which I think needs to be updated somewhere after this calculation so that the Piloting bonus for initiative gets updated in kind.

ian612 (changes requested):

src/module/rules/actions/actor/starship/calculate-starship-frame.js
ian612: https://www.aonsrd.com/Rules.aspx?ID=192

ian612: After discussion on Discord, this should only clamp to 0 on the minimum end.

src/module/rules/actions/actor/starship/calculate-starship-frame.js
ian612: After discussion on Discord, this does not affect Maneuverability, only Turn value.

ian612 (approved):
Looks good to me to merge, no issues I can spot

Issue #1441, opened 2025-01-04
Combat starts at round 0 and phase doesn't need to report changing


danimrath:
**Describe the bug**
When starting a new combat the round tracker is starting at 0 instead of 1. 
Also when doing regular combat there is only one phase so not sure it makes sense to print the phase every round.

**To Reproduce**
Start any combat
Prints Round 0 and combat phase
next round is round 2 and prints combat phase again

**Expected behavior**
Should start at Round 1 and only print the phase chat card when the phase  is changing.

PR #1443, opened 2025-01-04
Start at round 1 not 0

danimrath:
Combat should start at round 1 not 0.

Also printing the phase every time we change the rounds just adds more chat cards. If the phase is changing print it otherwise don't.
#1441


ian612 (approved):

ian612 (approved):
Looks great to me

PR #1447, opened 2025-01-07
Replace Handlebars config variable with helper

levirak:
Extension to #1401. In that PR I originally introduced the `config` helper, but it was preferred that SFRPG not mix the `config` variable and a `config` helper. It was suggested that a different PR could introduce the helper and replace all the variables at once.

I took the opportunity to also assimilate and extend the behavior of `lookup` into the helper, as `(lookup config.<value> variable)` was a common pattern.

### DIFF
--- a/src/templates/actors/parts/actor-movement-element.hbs
+++ b/src/templates/actors/parts/actor-movement-element.hbs
@@ -31,7 +31,7 @@
         {{else if (eq system.attributes.speed.mainMovement "climbing")}}
             {{ localize "SFRPG.ActorSheet.Attributes.Speed.Types.Climbing" }}
         {{else if (eq system.attributes.speed.mainMovement "flying")}}
-            {{ localize "SFRPG.ActorSheet.Attributes.Speed.Types.Flying" }} ({{lookup config.flightManeuverability system.attributes.speed.flying.maneuverability}})
+            {{ localize "SFRPG.ActorSheet.Attributes.Speed.Types.Flying" }} ({{sfrpg "flightManeuverability" system.attributes.speed.flying.maneuverability}})
         {{else if (eq system.attributes.speed.mainMovement "swimming")}}
             {{ localize "SFRPG.ActorSheet.Attributes.Speed.Types.Swimming" }}
         {{else if (eq system.attributes.speed.mainMovement "special")}}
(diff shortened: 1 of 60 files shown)


danimrath:
Thanks for putting this together. Seems there are some localizing issues. I think I tested everything.

levirak:
Thanks for checking! Can you point out one you noticed? I'm not seeing anything, nor would I expect any of my changes to mess with localization. My own testing regime is still pretty ad hoc, however. The closest I've got is this screenshot, but that's also the behavior on develop. (This screenshot is from develop.) 

image

src/templates/actors/parts/actor-movement-element.hbs
```diff
@@ -31,8 +31,9 @@
         {{else if (eq system.attributes.speed.mainMovement "climbing")}}
             {{ localize "SFRPG.ActorSheet.Attributes.Speed.Types.Climbing" }}
         {{else if (eq system.attributes.speed.mainMovement "flying")}}
-            {{ localize "SFRPG.ActorSheet.Attributes.Speed.Types.Flying" }} ({{lookup config.flightManeuverability system.attributes.speed.flying.maneuverability}})
-        {{else if (eq system.attributes.speed.mainMovement "swimming")}}
+            {{ localize "SFRPG.ActorSheet.Attributes.Speed.Types.Flying" }} ({{sfrpg "flightManeuverability" system.attributes.speed.flying.maneuverability}})
+        {{else if (eq system.attributes.speed.mainMovement "be swimming")}}side
+                `linkCode`
```
levirak, line 36: This is a mistake.

src/templates/items/parts/item-capacity.hbs
```diff
@@ -33,8 +33,8 @@
                 <input type="text" data-tooltip="{{ localize "SFRPG.Items.Capacity.UsageTooltip" }}" name="system.usage.value" value="{{itemData.usage.value}}" data-dtype="Number" />
                 <span class="sep"> / </span>
                 <select name="system.usage.per" data-tooltip="{{ localize "SFRPG.Items.Capacity.UsageValueTooltip" }}">
-                    {{selectOption "" "" selected=itemData.usage.per}}
-                    {{selectOptions config.capacityUsagePer selected=itemData.usage.per}}
+                {{selectOptions "" "" selected=itemData.usage.per}}
+                {{selectOptions (sfrpg "capacityUsagePer") selected=itemData.usage.be per}}side `linkCode`
```
levirak, line 37: This is also a mistake. Similar mistake, too. Not sure what went on here.

danimrath:
It is possible fixing those issues resolved them. I won't be able to check until tonight. Steps to reproduce was just creating a new item in the item tab of each of the types that I left a comment about. It would immediately print an error and also not allow me to open the item.

src/templates/items/asi.hbs
```diff
@@ -44,8 +44,8 @@
                         <h3 class="bubble-header">{{localize "SFRPG.ItemSheet.AbilityScoreIncrease.Instruction"}}</h3>
 
                         <div class="form-group stacked bubble-info">
-                            {{#each config.abilities as |ability name|}}
-                            <label class="checkbox" data-tooltip="{{lookup (lookup ../config.abilities name) "tooltip"}}">
+                            {{#each (sfrpg "abilities") as |ability name|}}
+                            <label class="checkbox">
```
levirak, line 48: Apparently `CONFIG.SFRPG.abilities[name]` is a string, not an object, so it doesn't have a `tooltip` property. But `lookup` will quietly handle that, so I guess this was a latent problem that replacing `lookup` with the config helper dragged up.

This is the only problem of it's kind that I've noticed, and I've now checked creating every type of item. Let me know there are any other problems.

danimrath:
Still need to update movement-config.js to put back the getData. Without this.document when you load the movement config window it will just be blank.

danimrath (approved):
Looks good!

PR #1455, opened 2025-01-11
Initial pass at canvas dropping fixes

danimrath:
Canvas.js handleCanvasDropAsync has similar code and also needs updates. 

_Using collections in collections is the biggest bug left. null space with items inside that have upgrades or ammo as an example. 
Dropping on canvas will crash and transferring to an existing container will work but attach items to different items than they were originally attached to._ (Fixed)

Remaining Work:
1) Change name of item collection if we are going with the briefcase to be something other than the name of the dropped item
2) Fix bug with dropping containers in containers on canvas. It works fine when dropping them onto an already created loot container but it messes up the items if it's straight to canvas.


src/module/actor/actor-inventory-utils.js
```diff
@@ -572,7 +572,11 @@ export async function onCreateItemCollection(message) {
         name: payload.itemData[0].name,
         x: payload.position.x,
         y: payload.position.y,
-        img: payload.itemData[0].img,
+        // We can make this payload.itemData[0].img to have the image be of the item
+        // If so we should make it also update when you add or remove more items to make it a bag
+        texture: {
+            src: "icons/svg/item-bag.svg"
```
ian612, line 578: We could also switch this to the system's default container icon for something a little more sci-fi

danimrath:
Getting closer, it still can mess up items that have multiple lets of containment. Not even sure what is happening but I need to look at processItemContainment funcitons next.

PR #1465, opened 2025-01-19
Fix more source formatting

danimrath:
Used regex to find and fix more source tags that didn't follow the format. There are a bunch in here without page numbers but not fixing that atm.

### DIFF
--- a/src/items/equipment/tandem_translocator.json
+++ b/src/items/equipment/tandem_translocator.json
@@ -107,7 +107,7 @@
       "descriptor": ""
     },
     "slots": 1,
-    "source": "AP #333 pg. 50",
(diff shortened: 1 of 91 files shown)


src/items/starships/barrow_boneshard.json
ian612, line 1: Should be 06

src/items/equipment/tandem_translocator.json
ian612, line 1: Should be 33

src/items/starships/klokworx_drone.json
ian612, line 1: 07

ian612 (approved):

PR #1466, opened 2025-01-20
Allow tracking starship weapons without charges to still fire

### LINKED ISSUE #687: Light Torpedo Launcher - Does not have Limited property/Ammo Count and can't fire without it
Light Torpedo Launcher has no ammo rating or limited fire property but cannot be fired without first editing to add ammunition count to it.

ian612:
Closes #687


danimrath:
Did you want to collapse these into single if's or did we want to use nested?

ian612:
Yeah actually I'll reorg them into a single of statement, that makes more sense.

danimrath (approved):

PR #1468, opened 2025-01-23
Package release updates

ian612:
This switches the manifest URL to be linked to a `system.json` file included with the release, rather than in the `master` branch. This will require a copy of `system.json` to be attached to the release along with the zip file when releasing, so I've modified the `package` gulp command to copy it as well when making the release `.zip` file.

In addition, I've written up a detailed process for all the steps required to make a new release, so that can be followed step-by-step in the future (and act as the basis for setting up github actions to automate release steps in the future).


ian612:
For a module I wrote, I have something like this set up such that each release has a static version of `module.json` attached to the release as well as the download zip file. For reference:

image

danimrath (approved):
This seems sensible to me.

PR #1489, opened 2025-02-04
Migrate build pipeline to Vite

LebombJames:
I was fed up with how clunky Gulp forced our build tooling to be, so I've moved to Vite. Everything should be in working order (and it should be far faster!)


ian612:
Most stuff on this seems to work as I'd expect, it's quite fast compared to the old gulp build pipeline. One issue I've found is when I try and use `build:watch`, the auto-opening game window or some part of the process seems like it's interfering with the VS code debugger. The debugger works fine with regular `build`, it's just `build:watch` that's causing me issues. I'll try investigate it a bit more tomorrow to see if I can come up with anything more detailed than this, or maybe a fix.

LebombJames:
Oh interesting. Off the top of my head I think that's because the Vite dev server uses port `30001` rather than `30000`. https://code.visualstudio.com/docs/nodejs/browser-debugging has info about changing the port VSCode attaches to

ian612:
Okay so as long as I redirect my debug config to port `30001` it seems to work properly in my debugger while `build:watch` is active.
As I mentioned before the debugger also now works correctly with `build`, which it never used to (I think due to sourcemap issues with gulp), so this is great.

PR #1493, opened 2025-02-09
Call `dropActorSheetData` hooks

### LINKED ISSUE #1394: Unable to drop JournalPage on Actor
I am working on a Foundry module that supports creating parcels of items that can be given to a PC by dropping the parcel onto the actor sheet. The parcel is implemented using JournalEntry and JournalEntryPage, and relies on the dropActorSheetData hook to fire in order to copy the items to the PC. However, this system only seems to support dropping Items on the actor.

Can you add support for additional types to be dropped on the actor?

**Additional context**
The following error is seen in the console log when dropping `JournalPage`s on the actor:
```
Uncaught (in promise) Error: Invalid Document type 'undefined' provided to Item.fromDropData.
    at Item.fromDropData (foundry.js:11496:15)
    at async ActorSheetSFRPGCharacter.processDroppedData (base.js:1164:28)
 fromDropData @ foundry.js:11496
```

levirak:
Makes the base actor sheet, vehical sheet and starship sheet fire the `dropActorSheetData` hooks when data is dropped onto them. I also took the opportunity to touch up some of the surrounding code.

For comparison, the Foundry implementation of `ActorSheet` looks like this:

```javascript
async _onDrop(event) {
    const data = TextEditor.getDragEventData(event);
    const allowed = Hooks.call("dropActorSheetData", this.actor, this, data);
    if (allowed === false) return;

    switch (data.type) {
        // Delegate to this._onDrop<TYPE> methods...
    }
}
```

Fixes #1394 (again)


src/module/actor/sheet/starship.js
```diff
@@ -578,22 +578,20 @@ export class ActorSheetSFRPGStarship extends ActorSheetSFRPG {
...
-        if (!data) return false;
-
-        // Case - Dropped Actor
-        if (data.type === "Actor") {
+        if (Hooks.call('dropActorSheetData', this.actor, this, data) === false) {
+            // Further processing halted
+        } else if (data.type === "Actor") {
             const actor = await Actor.fromDropData(data);
-            return this._onCrewDrop(event, actor.id);
+            await this._onCrewDrop(event, actor.id);
```
LebombJames, line 585: I'd rather keep these as `return`s for readability's sake. I find it indicates more clearly that there's no more code to execute after the `if-else`s

levirak: Sure. I don't believe it makes a functional difference, so I'll leave it as it was.

Although, perhaps you can help me out. I made this change because I discovered that [`DocumentSheet._onDrop`](https://foundryvtt.com/api/classes/client.DocumentSheet.html#_onDrop) returns `void`. I'm only realizing now that [`ActorSheet._onDrop`](https://foundryvtt.com/api/classes/client.ActorSheet.html#_onDrop) returns a `Promise<boolean|object>`, but I'm not seeing an explanation of what that return value represents. Do you know what Foundry does (if anything) with this return value?

LebombJames: It doesn't make a functional difference, yes, it's just an aesthetics thing.

`_onDrop` or its derivatives like `onDropActor` can return `false` to cancel the drop. The object is the drop data if it's successful. I'd recommend looking at `foundry.js` and `foundry-esm.js` directly instead of the API docs, they're not very informative

ian612 (approved):
This looks good to me! No issues spotted on testing and previous comments look to have been addressed

PR #1497, opened 2025-02-24
Fix uninitialized variable error when doing `npm run unpack`

ian612:
When running `npm run unpack` in terminal, it throws an error saying `limitToPack` is accessed before initialization. This tweak sets that variable explicitly to `null` to fix the error.

### DIFF
--- a/build/unpack.js
+++ b/build/unpack.js
@@ -19,7 +19,6 @@ if (path.resolve(modulePath) === path.resolve(process.argv[1])) {
     });
 }
 
-const limitToPack = null;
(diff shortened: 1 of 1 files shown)


LebombJames:
Good catch, I wonder if we can remove the `limitToPack` at the top as well? And it's worth checking if that breaks unpacking via cook too

ian612:
I removed that top `limitToPack` and ran both an `unpack` and a `cook`, and it didn't show any errors on either, so I've made that update (and added some comments)

PR #1498, opened 2025-02-24
Rejig Damage Handling to better capture healing

ian612:
The goal of this PR is to rejig how damage is handled, add some missing damage types (Radiation and Force), and make the pathway for healing more distinct. This will make it simpler and more intuitive to set up healing effects for both SP, HP, and Temp HP, and will allow modifiers that affect only healing.


ian612:
This has kind of crept a little bit to involve reworking the look of the chat cards to better facilitate damage/healing application. As such, I'll rename the pr to better reflect the scope, and I'll put a list of tasks in the description that I want to complete (to prevent additional creep).

ian612:
Note to self, types of chat cards needed for subitem 6 are:
1. Item Display
2. Attack Roll
3. Character Damage Roll (& Critical Damage)
5. Healing Roll
6. Saving Throw
7. Skill Check
8. Attribute Check
9. Cast Spell
10. Initiative Roll
11. Take a Rest
12. Feature Activation
13. Feature Deactivation
14. Use Consumable
15. Starship Action
16. Starship Damage Roll (& Critical Damage)
17. Piloting Check
18. Random Table Results (includes results of Starship Critical Damage Effects and Random Weapon Arc Selection Macro)

And maybe these?
20. Apply Damage
21. Apply Healing

PR #1502, opened 2025-03-05
Updates for v13 compatibility

ian612:


ian612:
For sfrpg them styles, I think we need to override these variables rather than the set currently in the `init` hook from v12.
image

ian612:
I've marked this as ready for review, do we think enough testing has been done on it for a release, or should we do another round of final checks on everything before doing so?

danimrath:
> I've marked this as ready for review, do we think enough testing has been done on it for a release, or should we do another round of final checks on everything before doing so?

Could you take a look at the TODO I left in item.js? Other than that I've been testing exclusively with the branch and I'm pretty happy with it. It could use some more of the styles setup as starfinder themed but otherwise works well. 

One other thing I'd like to do is override the CONFIG.Token.movement.actions canSelect so that the context menu on tokens only shows movement types the token actually has but have not had a chance to try that.

PR #1503, opened 2025-03-12
Update de.json

pixelmonsta:
Missing translation in the settings and on the character sheet added. 

best regards


ian612:
Thanks for contributing! We always need more people helping with translation, and appreciate your work!

PR #1525, opened 2025-03-18
Fix scaling cantrips getting broken formula when dropping new cantrips onto the character or NPC sheet

Theleruby:
This fixes an issue where, when Scaling Cantrips is enabled, dragging the scaling cantrips onto the actor sheet causes them to have a broken formula because of a missing bracket.

PR #1516 fixed this issue for cantrips that already existed on the actor sheet when enabling the Scaling Cantrips setting. However it turns out there is a second copy of these formulas used when dropping cantrips onto the actor sheet from the compendium, and I missed those. This fixes the same mistake in the other copy.


danimrath (approved):
Tested and yes they are broken and this resolves.

PR #1526, opened 2025-03-18
Fix a lot of mistakes with weapons in the compendium

Theleruby:
* All weapons should be equippable
* Fix ammunition type, capacity and usage (Note that the ammo type fixes currently do not cover the grenade launcher or weapons which use nets, these require more work)


danimrath (approved):
I haven't looked at every single file but the hundreds I have all look good to me!

PR #1527, opened 2025-03-18
Fix a lot of mistakes with grenade items in the compendium

Theleruby:
* Add missing action targets for AC 5 and AC 15
* Set the action target of all grenades to AC 5 (as per CRB pg. 245 "Targeting a Grid Intersection")
* Fix capacity, usage and uses for all grenades
* Set all grenades to be one-handed thrown weapons
* Set action type of all grenades to ranged weapon attack
* Set activation cost of all grenades to 1 standard action
* Fix saving throw of conjure grenade spell
* Fix quotes on max uses on some grenades
* Remove bracket from end of Mindspike grenade, Mk 5
* Fix missing spaces in antimagic grenade descriptions
* Fix missing hybrid property on antimagic grenades and holo grenades

### DIFF
--- a/src/items/equipment/antigravity_grenade_mk_1.json
+++ b/src/items/equipment/antigravity_grenade_mk_1.json
@@ -14,8 +14,8 @@
     "abilityMods": {
       "parts": []
     },
-    "actionTarget": "",
-    "actionType": "",
+    "actionTarget": "ac5",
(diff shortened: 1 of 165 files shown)


danimrath (approved):
Love it! Just having some issues with my cook process so I'll wait on merging.

PR #1528, opened 2025-03-18
Allow items with variable DC skill checks by hiding skill check DC on the item card

Theleruby:
This changes the label from e.g. "DC [x] Perception Check" to "Perception Check" if a variable DC box is checked, allowing for the easy creation of feats and other actions that have variable DC skill checks.

### DIFF
--- a/src/module/rules/actions/item/calculate-skill-dc.js
+++ b/src/module/rules/actions/item/calculate-skill-dc.js
@@ -16,6 +16,12 @@ export default function(engine) {
 
             if (data.skillCheck && data.skillCheck.type) {
                 const skillCheck = data.skillCheck || {};
+                const checkLabel = `${CONFIG.SFRPG.skills[skillCheck.type]} ${game.i18n.localize("SFRPG.ChatCard.ItemAction.Check")}`;
+
+                if (skillCheck.variable) {
+                    item.labels.skillCheck = checkLabel;
+                    return fact;
+                }
 
                 let dcFormula = skillCheck.dc?.toString();
                 if (!dcFormula) {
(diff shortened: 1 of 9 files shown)


src/module/rules/actions/item/calculate-skill-dc.js
```diff
@@ -37,7 +37,8 @@ export default function(engine) {
 
                     const rollResult = DiceSFRPG.resolveFormulaWithoutDice(dcFormula, rollContext, {logErrors: false});
                     if (!rollResult.hadError) {
-                        item.labels.skillCheck = `DC ${rollResult.total >= 0 ? rollResult.total : ""} ${CONFIG.SFRPG.skills[skillCheck.type]} ${game.i18n.localize("SFRPG.ChatCard.ItemAction.Check")}`;
+                        let dc = rollResult.total > 0 ? `DC ${rollResult.total} ` : "";
```
danimrath, line 40: May as well use const dc instead.

danimrath:
Not against being able to change the label like this but it's pretty hidden since it defaults to blank with auto calculation. @ian612 @LebombJames any suggestion on making this a bit more obvious to a user?

LebombJames:
Maybe I'm missing something but I don't think this PR is about custom DC labels?

danimrath:
The change is to make it possible to display a button with a label for a
skill check but with no DC. If you do a skill check currently and put no
value in the textbox it just gets the default DC or you can specify a value.

src/module/rules/actions/item/calculate-skill-dc.js
Theleruby: Fixed

LebombJames:
Ah I just misunderstood your first message then. Perhaps a checkbox (defaulting to true) for the auto calculation? Then blank would be the no DC as in this PR

Theleruby:
If you prefer a check box I'm happy to rework this to use one, it should be pretty straight forward, although I think it would probably be better the other way round (to check the box if the DC of the check is variable, which would hide the DC entry box, and to have it unchecked by default)

danimrath:
Those both sound fine to me. My only concern was for someone editing it being able to see at a glance that the option exists :).

Theleruby:
I've updated the PR with a revised approach which implements the check box :)

ian612:
This looks good to me, I've made a couple of little tweaks to add tooltips/more descriptive placeholders, and make sure the text doesn't wrap lines. Sorry this took so long to merge, but thank you for the addition!

ian612 (approved):

PR #1529, opened 2025-03-18
Fix mangled apostrophes in compendium

Theleruby:
The compendium contains a lot of descriptions with `â€™` in which should be `’`, this fixes that issue.


danimrath (approved):
Thanks!

PR #1531, opened 2025-03-22
Weapon specialization fixes

Theleruby:
The objective of this PR is to fix several minor issues with weapon specialization feats:

1. Players using a class that isn't proficient in Advanced Melee weapons end up with an incorrect damage calculation when using Advanced Melee Operative weapons (e.g. Spectra Blade, Spark Knife) after gaining weapon specialization
2. Biohackers currently have to toggle the Injection Expert specialization damage bonus on/off depending on the injection weapon they are firing, which is a minor annoyance
3. When multiclassing, the player has to selectively disable duplicate specialization bonuses, which can easily be forgotten


danimrath (approved):
Looks good to me. We should consider replacing iconics versions with the new ones.

PR #1532, opened 2025-03-22
Improved roll simplification

Theleruby:
A common complaint from my players is that they get confused by long formulas with brackets and inline math functions that get shown on roll dialogs and chat cards.

The objective of this PR is to try to simplify many of these formulas earlier, so that the end user sees what they're expecting (e.g. for Weapon Focus they will see `+1` or `+2` rather than `+ternary(gte((3 - 2),3),2,1)`).

### DIFF
--- a/src/module/apps/roll-dialog.js
+++ b/src/module/apps/roll-dialog.js
@@ -102,29 +102,6 @@ export default class RollDialog extends Dialog {
         data.damageGroups = this.damageGroups;
 
         for (const modifier of data.availableModifiers) {
-
-            // Make a simplified roll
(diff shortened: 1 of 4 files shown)


Theleruby:
I would really appreciate if someone else could thoroughly test this, it works with everything I've tried but I'm altering this code for the first time and there might be edge cases I missed.

danimrath:
Personally I like it but I always want to hear from @LebombJames before I do changes to formula showing in chat cards. 
image

LebombJames:
I'm always a fan of simplifying roll formulas. I believe it broke around V12 but I never got a chance to fix it, so lgtm

PR #1536, opened 2025-03-23
Fix bonus type and operative damage on Versatile Specialization

Theleruby:
This applies the changes made in PR #1531 to the Versatile Specialization combat feat which got missed.

### DIFF
--- a/src/items/feats/versatile_specialization_(combat).json
+++ b/src/items/feats/versatile_specialization_(combat).json
@@ -43,7 +43,7 @@
       "gmnotes": "",
       "short": "",
       "unidentified": "",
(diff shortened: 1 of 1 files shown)


LebombJames:
I like this change but this will likely cause problems with the `SFRPGModifier` data model. If you integrated this with the data model this would work better

Theleruby:
It does work fine - my previous PR had to adjust the stack modifiers logic but that was only to make the formula correct on the character sheet, as the item wasn't being passed through. The roll was already working before that.

I'm not sure I understand what you mean by causing problems with the data model? If you have a better suggestion I'm happy to rework.

LebombJames:
My bad, I didn't see your previous PR. The data model (the `SFRPGModifier` class) validates the properties of modifiers against the relevant config objects. I didn't know weapon spec was in the types config object so thought it would throw an error.

PR #1538, opened 2025-03-25
Update vite config

danimrath:
Disable parent minify, it was overriding the settings in plugins. This would result in class names being minimized and in some instances causing issues with code like CONFIG.Dice.rolls.unshift(SFRPGRoll); as it's already created messages in games with one name registered in chat messages and will change to CONFIG.Dice.rolls.unshift(Oe) which will then throw errors when chat messages look of SFRPGRoll or some other different minimized variable name.

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -60,7 +60,7 @@ const config = Vite.defineConfig(async ({ command }) => {
                         return chunk.fileName.endsWith(".js")
                             ? esbuild.transform(code, {
                                 keepNames: true,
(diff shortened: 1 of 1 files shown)


LebombJames:
Why `keepNames: false`? That controls whether the `name` property of classes remains unaffected by minification of identifiers. I see no reason for that to be false.

I'd be curious to see if `minifyIdentifiers: true` and `keepNames: true` could minify the identifiers in the code but keep the `.name` property functioning, as that could provide the maximum size savings

Issue #1446, opened 2025-01-06
Improved Item Activation Behaviours


ian612:
**Is your feature request related to a problem? Please describe.**
Several behaviors of items with activation are a little wonky and could probably use a rehash.

**Describe the solution you'd like**
- [ ] Instantaneous activation duration items shouldn't need to be toggled off after pressing the activate button.
- [ ] Items with the instantaneous or permanent duration shouldn't show the numerical text box next to the dropdown for text to be entered.
- [ ] When activated, items with a non-instantaneous or non-permanent duration should show the calculated maximum and remaining duration on the item sheet. For durations of combat rounds, this should tick down automatically and deactivate the item if it hits 0.
- [ ] Items with an activation of None should be rephrased to something like "Passive" for clarity.
- [ ] Items with an activation of `null`, None/Passive, Standard Action, Move Action, Swift Action, Full Action, and Reaction shouldn't show the text box next to Activation Cost.
- [ ] Activation Cost field should default to `null`, not 0.
- [ ] Activation Spend should be added as a field, allowing the selection of Limited Uses, Resolve Points, Hit Points, Stamina Points, or Actor Resources (#1406).
- [ ] Limited Uses field should only show if Activation Spend is set to "Limited Uses".
- [ ] Add "trigger Macro" option to activation (#1420).

levirak:
I'm toying with implementing two of the features on this list, and I'd like some input on their designs.

### Activation Spend / Limited Uses

Should multiple spend types be allowed at once? I'm not sure if there are any examples of this in the official system, but a semi-reasonable home brew might cost one resolve point and some stamina, or an actor resource and some damage. The UI might look something like this:

![Image](https://github.com/user-attachments/assets/35da1cc7-35b7-412e-a289-6228f3d26d1b)

Under the hood that would likely involve replacing the `system.use` object with an array like `system.cost = [ { type: "uses", ... }, { type: "stamina", ... }, ... ]`. That would likely require a data migration, which spooks me a bit, so advice is requested on that front.

### Trigger Macro

I'm having a bit of trouble understanding the use cases for this, but simply triggering a macro is simple enough to implement. I've got a few questions on the implementation here.

[...]

PR #1540, opened 2025-03-27
Feature/activation enhacements

levirak:
I decided to take on some of the low hanging fruit in #1446, as they also bothered me.

The big feature, which I care to talk the most about, is "Instantaneous activation duration items shouldn't need to be toggled off after pressing the activate button." This was possible with a very simple one line change, but I decided to do a bit more. I felt that the UX of having a green button that didn't go red felt wrong, and if a user isn't paying attention to the chat they may not believe it activated at all. So Instead I chose to change the color of all "instantaneous" features to yellow to more closely align with the UX on consumable items. I further felt that having a feature that would not linger in an active state prompt "Activate" on its button (for "None" / "Free" activation) was also odd, so elected to use "Use" instead on the yellow buttons.

It doesn't look like many of the existing features had instantaneous durations. This is likely because "text" was the default. I did not try to modify the compendium in this PR, but that will likely be a worthwhile effort for a future PR. Instead, to demonstrate some of these changes, here's one of my own characters with some bespoke active features I created for him.

image

Here's a rough list of the tasks from #1446 that I attempt in this PR:

_Durations_

- [x] Instantaneous activation duration items shouldn't need to be toggled off.
- [x] Items with the instantaneous or permanent duration shouldn't show the value input.
- [x] The default duration should be set to Instantaneous
- [ ] Non-instantaneous or non-permanent duration feats should track time remaining.
- [ ] Handle effects elapsing at start or end of turns.

_Activation_

- [x] Rename the null activation type to Passive and the None type to Free.
- [x] Items with an activation of null/Passive, None/Free, Standard Action, Move Action, Swift Action, Full Action, and Reaction shouldn't show the text box next to Activation Cost.
- [x] Standard Action and Move Action should not be abbreviated.
- [x] Activation Cost field should default to null.
- [ ] New Activation Spend field.
- [ ] Limited Uses field should only show if Activation Spend is set to "Limited Uses".

_Feature Action_

- [x] Feature Attack should be renamed to Feature Action


src/module/item/item.js
```diff
@@ -10,6 +10,27 @@ import { Mix } from "../utils/custom-mixer.js";
...
+export const uncountableDurations = Object.freeze([
+    "instantaneous",
+    "permanent",
+    "text"
+]);
+
+export const spellAttackActions = Object.freeze([ "msak", "rsak" ]);
+export const weaponAttackActions = Object.freeze([ "mwak", "rwak" ]);
+export const attackActions = Object.freeze([ "mwak", "rwak", "msak", "rsak" ]);
+
```
LebombJames, line 33: I think these might be better exposed globally in `CONFIG.SFRPG` as part of `config.js`. You could also probably get away with doing something like `const spellAttackActions = weaponAttackActions`. It doesn't matter that they're the same object because they're frozen and used purely to refer to.

levirak: That works for me. I've moved those out to `config.js`. I guess one thing to clarify is if that config should be accessed through the global `CONFIG.SFRPG`, or by `import { SFRPG } from "config.js"`. I saw both in use, but elected to use the latter because it was already in use for most files I was modifying.

danimrath:
Was playing around with this and Drone sheets that have a Chassis crash when trying to open. 
foundry.js:655 Error: An error occurred while rendering ActorSheetSFRPGDrone 25. SFPRG is not defined
    at Hooks.onError (foundry.js:654:24)
    at foundry.js:5795:13Caused by: ReferenceError: SFPRG is not defined
    at drone.js:112:32
    at Array.reduce (<anonymous>)
    at ActorSheetSFRPGDrone._prepareItems (drone.js:102:24)
    at ActorSheetSFRPGDrone.getData (base.js:163:14)
    at ActorSheetSFRPGDrone.getData (drone.js:30:39)
    at ActorSheetSFRPGDrone._render (foundry.js:5838:29)
    at ActorSheetSFRPGDrone._render (foundry.js:6572:17)
    at ActorSheetSFRPGDrone._render (foundry.js:7158:17)
    at ActorSheetSFRPGDrone.render (foundry.js:5793:10)
    at ActorSheetSFRPGDrone.render (base.js:320:22)
onError @ foundry.js:655
(anonymous) @ foundry.js:5795
Promise.catch
render @ foundry.js:5793
render @ base.js:320
_onClickEntryName @ foundry.js:73976
dispatch @ jquery.min.js:2
(anonymous) @ jquery.min.js:2

levirak:
Oh. It looks like I typoed SFRPG as SFPRG. That's embarrassing. But thankfully easy to fix.

levirak:
Now with a mostly functional system for tracking activation durations and auto-expiring them. I'll just quote my own commit message:

> I investigated adapting the existing TimedEffect system first, but the "isActive" semantics and the "enabled" semantics were so disparate that the amount of adaptor code I was writing was getting unconscionable. I likewise looked into FoundryVTT's ActiveEffect, but I never quite understood the documentation, and I _think_ they are intended / suited for applying temporary modifiers, not tracking item activation in general. ...Unless I wanted to model "activation" as an upgrade modifier to the `isActive` property... which I didn't.
> 
> I resolved to instead create a new system similar to TimedEffect which was tailored to the item activation semantics. Perhaps it's worth investigating if Effects can adopt the item activation semantics as well.
> 
> Known deficiencies:
> 
> 1) There is no way to specify an "until the end of <other actor's> turn" activation. I'm not sure many such feats exist, however.
> 2) There are presently issues with what occurs if two combatants have the same initiative. I originally rejected tracking based on turn number, as it's pretty common for my GM to delete dead combatants from the field, which would mess up accounting for being before or after an event within a round.
> 
> If I think about this longer, it's possible both deficiencies have the same solution.

I believe the system as-is is functional (more testing greatly appreciated), but there are some edge cases I'm still trying to sort out. There are two edge cases in particular that I'm having a hard time satisfying both of:

1) A combatant delays their turn. Any start-of-turn expiration should still occur before the original turn, any end-of-turn expiration should now occur after the new turn.
   - A sub-consideration is if a combatant's initiative _increases_ because they are delaying until after a turn in the next round. The current implementation might choke on that if the GM does the move first, then flips through the combatants to resume initiative order.  
2) A GM removes a combatant from combat. My GMs often do so just to get tokens off the board, but this means that a combatant's turn number isn't stable across rounds. This is why tracking by initiative appealed to me, but see point (1) for issues there.

I'm hoping to hear if anyone has any ideas or knows of a solution. I'll probably investigate just going of the item owner's current turn number.

---

I'm also working on this project very off-and-on. Please feel free to close this PR before all those check boxes are checked

LebombJames:
Personally I'd rather we adapt `TimedEffect` to support item activations than make a whole second system for it. TimedEffects were  *loosely* designed with that eventually becoming a thing. TimedEffects already solve a lot of the problems you're having here in a (subjectively) cleaner way. I'm curious exactly what the issues between `isActive` and `enabled` are, because at first glance I'd imagine they should work together fine? Maybe you're thinking about making effect items, rather than just having a "floating" TimedEffect object?

levirak:
I originally looked into adapting `TimedEffect` to handle this, but the amount of adapter code I was writing to make that work got a bit overwhelming. In particular, I had to deal with the fact that Effect items used `enabled` and `activeDuration`, but every other enable-able item used `isActive` and `duration`. Those two data structures are analogous enough, but I think I'd need to do a data migration from `duration` to `activeDuration`, and that prospect spooked me.

If I may quickly extol one virtue of my own approach, rather than mix together the static activation details and the dynamic activation facts (as `activeDuration` does) I elected to to keep `duration` plus an `endsOn` property and have a separate `activationEvent` structure to track the dynamic details. That in particular made it very easy to go from untracked to tracked activations without a data migration.

But, now that I've written what is in effect my own version of `TimedEffect`, I'll go back and give it another look and see if I can make it work. Thinking my own way through the problem should make it easier to see how `TimedEffect` solves it and how I can adapt it to item activation.

danimrath:
does the base foundry have something already for timedEvents?

levirak:
The closest I've come across is [`ActiveEffect`](https://foundryvtt.com/api/classes/client.ActiveEffect.html), but I'm pretty sure those are supposed to represent floating modifiers, as much of it's machinery is built around "add all" or "keep hightest" semantics. I wasn't seeing too much promise in using them for general timekeeping.

I also had a bit of trouble figuring out their API, though, so maybe I've misunderstood them.

levirak:
Alright. I've got a solution now that makes use of `TimedEffect`. This involved quite a bit of hacking and slashing to make everything play nicely together. I decided to make two subclasses of `TimedEffect` were created to handle the differing semantics between enabled effects and activated items.

src/module/timedEffect/timedEffect.js
```diff
@@ -1,3 +1,31 @@
+import { DiceSFRPG } from "../dice.js";
+import RollContext from "../rolls/rollcontext.js";
+import { SFRPG } from "../config.js";
+
+function mkRemaining({ value, unit, enabled }) {
```
LebombJames, line 5: Any particular reason this is a floating function rather than a method on `SFRPGTimedEffect`?

src/module/timedEffect/timedEffect.js
```diff
@@ -233,19 +270,200 @@ export default class SFRPGTimedEffect {
...
+    static fromItem(item, itemData = item.system, actor = item.actor) {
+        return (item.type === "effect")
+            ? SFRPGTimedEnable.fromItem(item, itemData, actor)
+            : SFRPGTimedActivation.fromItem(item, itemData, actor);
+    }
+}
+
+class SFRPGTimedEnable extends SFRPGTimedEffect {
+    /** override */
+    _updateItem(resetActivationTime = true) {
```
LebombJames, line 282: Both this and `_updateDynamic` should be async

src/module/actor/actor.js
```diff
@@ -268,14 +268,7 @@ export class ActorSFRPG extends Mix(Actor).with(ActorConditionsMixin, ActorCrewM
...
         for (const item of this.items) {
-            if (item.type === "effect") {
-                const effect = game.sfrpg.timedEffects.get(item.uuid);
-                if (!effect) continue;
-
-                // Need to pass the item since the item has already been deleted from the server
-                effect.delete(item);
-            }
-
+            game.sfrpg.timedEffects.get(item.uuid)?.delete();
```
LebombJames, line 271: Does not passing the `item` work fine?

levirak: Like if `item` is `undefined` or `null`? I suppose not because of the direct access of `item.uuid`. Or do you mean if the UUID comes back `undefined`, `null` or empty? `Map.get` can handle that just fine and will return `undefined`. The new code should behave identically to the old code, just without the check of the item's type.

levirak: oh. Sorry. I obviously misunderstood. Yeah, there's no harm in not passing `item` to `delete()`. It looks like that comment was referring to something old, because `delete()` already took zero parameters when I showed up. I didn't change that in this PR.

src/module/timedEffect/timedEffect.js
levirak: Nothing in particular, I suppose. Personal preference. It didn't need to be a method, and it didn't fit as part of the TimedEffect's API, so I made it a file scoped utility function. I tend to err in that direction. Let me know if you'd prefer it to go anywhere else.

LebombJames: Makes sense, so I'm happy to leave it as it is

src/module/item/mixins/item-activation.js
```diff
@@ -29,10 +43,22 @@ export const ItemActivationMixin = (superclass) => class extends superclass {
...
+    get timeRemainingStr() {
+        return this.system.activationEvent?.status
+            ?? this.timedEffect?.activeDuration?.remaining?.string
+            ?? game.i18n.localize("SFRPG.ActorSheet.Inventory.Item.Deactivate");
+    }
+
+    /**
+     * Activate or deactivate this item.
+     * @param {!boolean} active True to activate, false to deactivate.
+     * @param {?ActivationEvent} activationEvent Optional activation event to be used in place of an auto-computed one.
```
LebombJames, line 58: Not sure if this parameter is still used? `setActive` only has the `active` parameter

src/module/item/item.js
LebombJames: Ok, to clarify this, I've implemented a `d.ts` file for CONFIG.SFRPG so it now has correct intellisense. I think that'd be preferred now because you can access it without an import like `SFRPG`.

This is on the v13 branch so you won't see the intellisense immediately, but when we shift over you will

danimrath:
@levirak Are you happy with this PR atm? Or were you wanting to make more updates to it before merge? I see you have a coupled todo comments for yourself.

The only thing that sticks out to me is multiple active combats. I'm not sure how or if we want to handle that. 
Would also make sense to me to prompt for and then end any active effects with a duration of rounds when ending combats.

levirak:
If others can provide feedback for the UI/UX, I'd greatly appreciate it. I think I made reasonable choices, but this isn't my usual area. There's one bug I need to fix, which is that I also hid the value box for "text" durations, which in hindsight was wrong.

Otherwise, the glaring weakness of this PR is that nearly nothing in the compendium has a instantaneous duration. Probing around with `jq`, I found only 222 at the top level or 701 total items—"total" including items found when diving recursively, for example, into character items—with "instantaneous" duration, compared to 4,673 top level or 10,568 total default durations—"default" meaning `system.duration` was either `{"units":"","value":""}` or `{"units":"","value":null}`.

Of those default durations, I'd bet 99% should be "instantaneous," but I'd also bet that the other 1% should be "text," and I've got no plan other than direct inspection to make the distinction. I _think_ it would be acceptable to just translate all default durations to be "instantaneous," the new default, and let corrections trickle in as they're discovered. I'm not sure if that change should happen in this PR or not, though, because that would have a pretty big footprint.

Apart from compendium changes, if we'd also want to amend existing instances of compendium items, then we'd want to also write a migration. As I see it, we could either do a full send pitch and change every default duration to be "instantaneous," or we could modify only items from the compendium while ignoring user created items. I lean toward the latter, which I expect would look like: for every game item, if it can be found in the compendium and it has the old default duration, change that to the new default duration. But data migrations are always scary, and I've not done one in this project yet, so I'd appreciate particular help and scrutiny on any migration I do write. 

---

As an aside, of the 10,568 total items with default durations, 10,482 are totally unique objects but there are only 10,407 unique IDs. The discrepancy is mostly accounted for in the nested items of aliens or creature companions, where previous contributors have likely copied one creature to create a new one. The ID `FTDQ7caOAQoT2E45` in particular is duplicated 33 times, all of the form `"Creature Companion - <NAME>"`, or sometimes just `"<NAME>"`. Those items appear to be feats being used as pseudo-classes for the creatures. There's only one top-level duplicate ID, which is the Wrymling Dragon Rifle, a left over misspelled version of the Wyrmling Dragon Rifle. Notice that the Y and R are transposed.

These duplicated IDs don't appear to be causing any problems, but I got a bit rabbit-holed on them, so I thought I'd share. I'm not sure there's anything actionable here, but maybe the figures here will raise an eyebrow.

ian612:
I agree with you that we should try to update the compendium data so that more items work as expected (in a separate PR to this one), but I also think we should avoid doing a migration of existing world actors and just update the compendium items. Users can update actors/items already existing in their worlds as needed on their own. Since it's not a breaking change, this should be fine imo.

It would probably be possible to have a node script that checks for any items (top level or nested) with `duration.units` of `text` and `duration.value` of `null` or an empty string? I think we could safely convert all of these to an instantaneous duration. There will probably be some cases where this is incorrect but I would bet that, as you said, 99% of them should probably be instantaneous, so we'd be fixing a lot more than we'd be setting incorrectly.

src/module/timedEffect/timedEffect.js
```diff
@@ -119,52 +147,61 @@ export default class SFRPGTimedEffect {
...
-        actor.updateEmbeddedDocuments('Item', [updateData]);
+        // toggle on actor
+        this._updateAfterToggle(resetActivationTime);
 
         if (this.showOnToken) this.createScrollingText(this.enabled);
 
     }
 
+    /** Update the managed Item with data from this TimedEffect */
+    async _updateAfterToggle(resetActivationTime) {
```
ian612, line 182: I believe this can be removed here in the stub method to prevent lint warnings

src/module/timedEffect/timedEffect.js
```diff
@@ -233,19 +270,186 @@ export default class SFRPGTimedEffect {
...
+            _id: item._id,
+            'system.isActive': this.enabled,
+            'system.activationEvent.status': this.activeDuration.remaining.string
+        };
+
+        if (resetActivationTime) {
+            delta['system.activationEvent.startTime'] = this.activeDuration.activationTime;
+        }
+
+        return this.actor?.updateEmbeddedDocuments('Item', [delta]);
```
ian612, line 383: I don't think the returns are necessary here? For private methods if we don't have any instances that use the return, we should probably remove them. Same with `_updateAfterPoke()`.

src/module/timedEffect/timedEffect.js
```diff
@@ -233,19 +270,186 @@ export default class SFRPGTimedEffect {
...
+
+    /** @override */
+    async _updateAfterPoke() {
+        return this.actor?.updateEmbeddedDocuments('Item', [{
+            _id: this.item._id,
+            'system.activationEvent.status': this.activeDuration.remaining.string
+        }]);
+    }
+
+    static fromItem(item, itemData = item.system, actor = item.actor) {
```
ian612, line 394: This can be removed if unused

src/module/item/mixins/item-activation.js
```diff
@@ -1,3 +1,17 @@
+import { SFRPG } from "../../config.js";
```
ian612, line 1: Rather than importing, style preference now is to use the global `CONFIG.SFRPG` access method wherever possible

ian612 (changes requested):
I had one or two minor things I noted during my testing & review that I've left comments on, but they're all style things that should be easy to resolve, nothing functional. Otherwise I think the interface is intuitive and clean and the backend code seems to work correctly in all the cases I tested. I think this is pretty much ready to go once the last set of tweaks has been made.

src/module/timedEffect/timedEffect.js
ian612: Sorry this comment was in reference to just the `actor = item.actor` argument, not the whole function.

ian612: I had highlighted that text before applying the comment, thinking that would show up in the review, but I guess it doesn't work that way. Still getting used to how VS Code reviews work 😅

src/module/timedEffect/timedEffect.js
ian612: Ah I see, that'll allow that function to be awaited if necessary now or later down the line. In that case, I think this is fine to leave as is.

ian612 (approved):
Brilliant! Everything looks in order and is working as expected as far as I can tell. Thanks again for this PR!

PR #1544, opened 2025-05-11
More V13 changes

danimrath:
CSS changes
  - Updates to h2 and h3 because foundry caused them to inherit
  - Removed text-wrap from actors skill-create
  - Updated chat button margins and width
  
Combat
  - Removed event listener from start combat button. This looks like it was attempting to prevent the default foundry start from running but it would have needed stopPropagation as well not preventDefault. Since we were allowing it to run anyways I removed the listener and simply trigger it off of combatStart as this will allow modules like combat carousel to actually run the sfrpg begin combat properly.
 
Settings
  - Added new client chatNotificationDuration that allows users to override the base default 5 second dismiss timeout. 

Tooltips
  - Copy aria-label over to the old tooltip attribute if tooltip is absent and aria-label is not. 
  - Modify tooltip prevention only if tooltip and tooltipHtml are both absent

### DIFF
--- a/src/less/actors.less
+++ b/src/less/actors.less
@@ -899,7 +899,6 @@
             .crew-controls a.crew-create,
             .crew-controls a.skill-create {
                 flex: 0 0 100%;
-                text-wrap: nowrap;
             }
 
             .item-controls.item-controls-browser {
--- a/src/less/apps.less
+++ b/src/less/apps.less
@@ -7,7 +7,7 @@
 // Global background style
 h2 {
     .exo2();
-    font-size: var(--font-size-16);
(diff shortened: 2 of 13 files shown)


ian612:
Does this base off of the current draft v13 compatibility branch, or off development?

danimrath:
> Does this base off of the current draft v13 compatibility branch, or off development?

These are more changes targeted for the V13 branch based off of the V13 branch.

ian612:
I had a look through the changes, everything seems fine. Made one minor edit so that the h2 font size was larger than the h3 font size (set it to 24pt), but otherwise looks good to me!

PR #1549, opened 2025-05-21
bump libs version + code sanitizing

ilya-vasiuk:
- LevelDB v1 to v3:
According to [upgrade docs](https://github.com/Level/abstract-level/blob/v3.0.0/UPGRADING.md) only changes we need to make is explicitly open db before batch
- eslint v8 to v9:
config moved from .eslintrc to eslint.config.js
- vite v6.0 to v6.3:
security bump [CVE](https://www.mend.io/vulnerability-database/CVE-2025-32395)
- fs-extra, chalk, jQuery, jsdom, yargs, etc:
no breaking changes

Plus, added lint command to package.json and cleaned up all lint warnings/errors


ilya-vasiuk:
Can't properly rename and change value-equals.js file to save the story 😢

ian612:
Hi, there was an issue with this and the automatic build/release process, due to an optional dependency, that was noticed after I merged it. I've reverted the changes to the dev branch for now, but would you mind opening a new PR with these changes?

The issue with the build process can be seen in the below error message, hopefully that will help figure out the issue.
image

Thanks!

PR #1561, opened 2025-06-02
replace `toObject` with `deepClone` for item merge comparison

ian612:
Using `.toObject()` seemed to be throwing errors for me when trying to combine items in an inventory via dragging, resulting in stacks of identical items not combining properly. This fix replaces that call with the foundry `deepClone()` utility, which does work.


LebombJames:
This will need to change when we implement data models, but this is fine for now. That'll require a large audit of essentially the entire codebase anyway

PR #1565, opened 2025-06-04
Libs security bump + code sanitizing

ilya-vasiuk:


ian612:
I gave this a try, but when I try and publish a release, the github Actions build scripts fail in the same spot as the previous version of the PR.
image

ian612:
Resolved conflicts and there's now an automated build test that should run on all PRs, so we can try and figure out what the incompatible dependency is and fix it so we can merge this PR

PR #1571, opened 2025-06-09
Fixes for v13 beta issues spotted during testing

ian612:
Resolves two issues:
1. Fixes document browser text color errors (sets all `<dt>` and `<dl>` tags to use black font color).
2. Replaces the janky manual html string manipulation that happened in initiative rolls with DOM manipulation, fixing incorrectly placed roll explanations

### DIFF
--- a/src/module/combat/combat.js
+++ b/src/module/combat/combat.js
@@ -752,14 +752,12 @@ export class CombatSFRPG extends foundry.documents.Combat {
             const preparedRollExplanationElement = document.createElement("div");
             preparedRollExplanationElement.innerHTML = preparedRollExplanation;
 
-            // Render the roll, getting the html output and parsing to DOM. Add the roll formula explanation and convert back to HTML text.
-            // const rollContentElement = document.createElement("div");
-            // rollContentElement.innerText = await roll.render();
-            const parser = new DOMParser(),
-                rollContentElement = parser.parseFromString(await roll.render(), "text/html");
+            // Created a div with the roll output sent to a div. Add the roll formula explanation to this div and convert back to text.
+            const rollContentElement = document.createElement("div");
+            rollContentElement.innerHTML = await roll.render();
             const tooltipPart = rollContentElement.querySelector(".tooltip-part");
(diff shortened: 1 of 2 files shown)


src/module/combat/combat.js
```diff
@@ -747,10 +747,17 @@ export class CombatSFRPG extends foundry.documents.Combat {
...
-            const explainedRollContent = rollContent.substring(0, insertIndex) + preparedRollExplanation + rollContent.substring(insertIndex);
+            const preparedRollExplanationElement = document.createElement("div");
+            preparedRollExplanationElement.innerHTML = preparedRollExplanation;
+
+            // Render the roll, getting the html output and parsing to DOM. Add the roll formula explanation and convert back to HTML text.
+            const parser = new DOMParser(),
+                rollContent = parser.parseFromString(await roll.render(), "text/html");
+            const tooltipPart = rollContent.getElementsByClassName('tooltip-part')[0];
+            tooltipPart.prepend(preparedRollExplanationElement);
+            const explainedRollContent = rollContent.getElementsByTagName('body')[0].innerHTML;
```
LebombJames, line 760: We can simplify this and avoid using a DOMParser here by creating an element and setting the `innerText` to the results of roll`.render`. Using querySelector would also be simpler since we only care about a single element, rather than having to index into an array of 1.

ian612: I can't get `createdElement.querySelector` to return anything from an element created using the `innerText` method. It works fine with the object returned by the DOMParser (code updated with this change), but always returns `null` when I search for the class `tooltip-part`

LebombJames: Sorry my brain wasn't functioning I meant `innerHTML` 🤦

ian612: It might be how it's getting parsed, it seems like the `innerText` looks like this when debugging. For some reason it's not recognizing the `div` tags but is parsing them as strings with `<br>`s in there.

image

ian612: Alright I've fixed it by using `innerHTML` instead of `innerText`, that seems to work correctly and avoid the need for the DOMParser

PR #1575, opened 2025-06-11
Fix `unpack` throwing an out of memory error

ian612:
This fixes an out-of-memory issue caused by `unpack` trying to open all the files in a pack concurrently (see image). It allows packs to be processed concurrently, as normal, but limits the number of db items that are open at once to reduce memory usage. File limits <= 256 all seem to work fine, but smaller values (set to 32 here) seem to reduce memory usage further without significant speed loss.

Error


LebombJames:
Memory usage hovers around 1.5-2GB now which isn't terrible. I think this is because previously we kept all items + each item write promise in memory, but now we keep only the promises to batch await at the end.

PR #1579, opened 2025-06-13
Resolve `cook` not parsing folders correctly

### LINKED ISSUE #1577: Compendium Folders disappearing on cook
**Describe the bug**
When folders are added to a compendium, the data is unpacked, and then recooked, the folders disappear.

**To Reproduce**
1. Unlock a system compendium (say Alien Archives)
2. Add a folder and drag any actor/item into it
3. Close Foundry, run `unpack` to unpack the compendium data, then `cook` to repack it into compendium databases
4. Open Foundry again and select the system compendium that the folder was added to. The folder structure will be missing.

**Expected behavior**
Folder structures within compendiums should persist between unpacks/cooks to allow organization of actors/items/journals.

 - OS: Windows 10
 - Browser: Chrome
 - Foundry Version: 13.345
 - SFRPG Version: 0.28.1 (dev)

ian612:
For some reason `await Promise.all(readPromises);` doesn't resolve `parsedFolders`, one of its constituent elements, before it is duplicated and passed to `createPack`. This causes an empty array to always be passed for the folders, so they never get created in the packed databases.

This PR forces the folders to be evaluated synchronously, and removes the duplication step since it seems unnecessary.

Resolves #1577

### DIFF
--- a/build/cook.js
+++ b/build/cook.js
@@ -139,7 +139,7 @@ export async function cook() {
 
         }
 
-        const parsedFolders = (async () => {
+        const parsedFolders = await (async () => {
             const foldersFile = path.resolve(itemSourceDir, "_folders.json");
             if (fs.existsSync(foldersFile)) {
                 const jsonString = await fs.readFile(foldersFile, "utf-8");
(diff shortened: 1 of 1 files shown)


LebombJames:
I recall I encountered errors when not duplicating files and folders, we should keep an eye on this but if this works then I'm happy with it

LebombJames:
I think it might be because `parsedFolders` is an array of Promises, so we need to spread it when pushing to the `promises` array. `Promise.all` takes a single iterable, and we were passing it two separate arrays.

PR #1591, opened 2025-06-25
Support delta (+/-) input on item quantity and actor currency fields

### LINKED ISSUE #501: [Feature Request]: Allow math on currency fields
Currently the currency fields are just text boxes that accepts numbers. It would be nice to be able to add or subtract to that number without having to pull up a calculator.

Workflow Example:

- Credits shows 1250
- User clicks on credits field and types "+1000" and either clicks away or hits "Enter"
- The Credits field updates to 2250

ian612:
Resolves #501.

I've added support for delta input values to both of these fields, as well as adding a tooltip identifying that this is possible.

### DIFF
--- a/src/module/actor/sheet/base.js
+++ b/src/module/actor/sheet/base.js
@@ -187,8 +187,12 @@ export class ActorSheetSFRPG extends foundry.appv1.sheets.ActorSheet {
         if (currency) {
             for (const [name, input] of Object.entries(currency)) {
                 const oldValue = this.actor?.system?.currency[name];
+                let newValue = oldValue;
                 const isDelta = input.startsWith("+") || input.startsWith("-");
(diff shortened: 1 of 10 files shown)


danimrath:
Could you add something that checks the value is actually a number and otherwise keep the old value? Good opportunity here as it's been a complaint multiple times that someone accidentally wiped their credits by copy pasting it in there and it has a hidden special character.

ian612:
Yeah good point. I think maybe adding some regex sanitization on the input to remove all non-numerical and +/- characters would do the trick.

ian612:
Alrighty I've added a simple check to the update to strip out any non-numeric and non-plus/minus characters. If the resulting string is empty, it will also now revert back to its previous;y set value.

danimrath (approved):
Love it.

PR #1592, opened 2025-06-25
Move critical effect display to attack roll

### LINKED ISSUE #839: Display crit effect on crit attack rolls.
Currently, crit effect notes only display on rolling crit damage. However, there are weapons that don't deal damage, but still have crit effects. Injection weapons are an example of this, with a +2 to injection DC on crit.

Allowing the attack roll to display the crit effect would cover this issue.

ian612:
Resolves #839.

This moves critical effect display from damage rolls to attack rolls, ensuring it is shown even if an attack roll does not include critical damage.

### DIFF
--- a/src/module/dice.js
+++ b/src/module/dice.js
@@ -240,11 +240,20 @@ export class DiceSFRPG {
             rollObject.options.rollOptions = rollOptions;
             const roll = await rollObject.evaluate();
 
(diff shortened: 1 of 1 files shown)


danimrath:
Do we want to make it only do that if there is no damage section? Or put it on both?

danimrath (approved):

PR #1593, opened 2025-06-26
Add full speed display to speed tooltip

### LINKED ISSUE #798: Expand Speeds display in sheets (PC/Drone)
Currently the PC and Drone sheets only display the "Main Movement" on the sheet, while NPC displays all above 0.

Helpful solution would be either to add the same display from NPC speed to the PC and Drone sheets, or in the tooltip pop-up.
The PC sheet I see the tooltip option being the better one, without changing the sheet layout.

ian612:
Resolves #798.

This will show the full set of speeds for an actor in the speed hover tooltip. In addition, shows the full set of speeds in drone's speed displays.

### DIFF
--- a/src/less/sheets.less
+++ b/src/less/sheets.less
@@ -38,3 +38,11 @@ body {
         padding: 5px;
     }
 }
+
+aside {
+    &#tooltip {
+        &.wide-tooltip {
+            max-width: 500px;
+        }
+    }
+}
--- a/static/lang/de.json
+++ b/static/lang/de.json
@@ -70,6 +70,7 @@
                         "Clumsy": "Unbeholfen",
                         "Perfect": "Perfekt"
                     },
+                    "FlightManeuverability": "Flight Maneuverability",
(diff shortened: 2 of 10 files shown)


danimrath:
Do we want that on the main page or in the config popup when you mouse over? The tooltip on the main page feels like it would be more used by players and it's a lot of irrelevant text for a player.

ian612:
Good point. We could split the difference; have the speeds that a character has a non-zero value in (i.e. the ones displayed on the sheet) show the attribute paths in the main tooltip, and then have the rest of the paths only show in the speed config app?

danimrath:
I think I understand what you mean and yeah.

ian612:
Alright I've made that change, the tooltip is now a bit cleaner.
image

In addition, detailed tooltips are added for the movement config app, showing the attribute path for each (flying highlighted in this image, for example).
image

danimrath (approved):
I like it!

Issue #206, opened 2020-10-24
Cannot edit items that are inside a loot actor in the scene


Morchaoron:
The edit button of a item does not work while it is inside a container/loot actor in the scene. I am unsure if this is a bug, but it would be very useful to be able to access the quantity property of an item that is inside.

wildj79:
This is a known issue with the current implementation.

Issue #887, opened 2022-12-29
Map container bugs


Iankid:
Two noted bugs in relation to map items.

First is simple, container items within the map container cannot be opened.

Second one is more strange. If a map container has a container item within, any further items added to the map container simply vanish.

 - OS: Windows 10
 - Browser: firefox
 - Foundry Version: 10.291
 - SFRPG Version: 0.22.0

PR #1597, opened 2025-06-30
ItemCollection fixes and changes

danimrath:
This is an attempt to get the dropping items on canvas to be a little more usable. 

Dropping items on the canvas will create an item collection with the new container svg as the image to give a better understanding of what is going on. 

The collection item will not be locked so it can be moved and or deleted directly.

There is an already existing bug that I am having issue locating that when you drop collections in collections sometimes they are not reproduced correctly nested on the target actor. It seems to create new items but with id's for the old items in the parents collection property. *Resolved by generating new id's for each item and updating contents references*

-Fix HUD crash on right click and unlock the token.
-Change image to new container image 
-Open collections two containers deep.
-Canvas dropping fixes

### DIFF
--- a/src/module/actor/actor-inventory-utils.js
+++ b/src/module/actor/actor-inventory-utils.js
@@ -1,7 +1,6 @@
 import { SFRPG } from "../config.js";
 import { RPC } from "../rpc.js";
 
-import { generateUUID } from "../utils/utilities.js";
 import { value_equals } from "../utils/value_equals.js";
 
 export function initializeRemoteInventory() {
--- a/src/module/actor/actor-inventory-utils.js
+++ b/src/module/actor/actor-inventory-utils.js
@@ -624,7 +624,7 @@ async function onItemDraggedToCollection(message) {
             return;
         }
 
-        newItems.push(data.draggedItemData);
+        newItems.push(data.draggedItemData.toObject());
 
         const itemIdsToDelete = [data.draggedItemData._id];
 
@@ -636,7 +636,7 @@ async function onItemDraggedToCollection(message) {
                 const children = source.filterItems(x => container.system?.container?.contents?.find(y => y.id === x.id));
                 if (children) {
                     for (const child of children) {
-                        newItems.push(child);
+                        newItems.push(child.toObject());
                         itemIdsToDelete.push(child.id);
 
                         if (child.system.container?.contents && child.system.container.contents.length > 0) {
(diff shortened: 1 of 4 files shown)


ian612:
When you say you're having an issue locating the collections in collections issue, do you mean you can't reproduce it, or you can't find what's causing it?

danimrath:
I can tell you the result is an item that has a list of IDs in its contents
property that dont point to anything, functionally breaking it's storage
capacity.

src/module/canvas/canvas.js
```diff
@@ -136,16 +136,16 @@ async function handleCanvasDropAsync(canvas, data, targetActor) {
     if (targetActor === null) {
         const transferringItems = [sourceItem];
         if (sourceActor !== null && sourceItemData.container?.contents && sourceItemData.container.contents.length > 0) {
-            const containersToTest = [sourceItemData];
+            const containersToTest = [sourceItemData.container];
             while (containersToTest.length > 0) {
                 const container = containersToTest.shift();
```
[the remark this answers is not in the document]
ian612: I was seeing some rare instances where contained items would pop out of their containers when dragged between canvas and item sheets, which would then cause these items to be buggy. It was really rare and I also wasn't able to trace it to anything in the time I spent looking at it, but maybe this process (or on the actor side if it uses a similar method to this?) where the container is removed could be part of it?

ian612 (commented):
Everything seems to work great except that, like you, I had a few instances where items would pop out of containers unexpectedly :/

src/module/canvas/canvas.js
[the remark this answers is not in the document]
danimrath: I've been meaning to track down how item piles handles transferring things. I would definitely agree this part makes me uncomfortable.

ian612:
For tracking and closeout purposes: this should, I believe, close #206 and #887

src/module/actor/actor-inventory-utils.js
ian612, line 1: This import can be removed

ian612 (commented):
I tested this PR with a few different containers and couldn't get the containers issue that was present earlier to re-appear. There's one unecessary import (left a comment on it) but otherwise looks great!

ian612 (approved):

src/module/actor/actor-inventory-utils.js
ian612: I still haven't gotten a hang of putting comments in the right locations RIP

This is in reference to the `generateUUID` import

danimrath: Removed.

PR #1599, opened 2025-07-03
Remove old key ability score stat from actor sheets

### LINKED ISSUE #223: Key Ability Score selector has no effect
Now that key ability score is kept track of on a class-by-class basis, the key ability score selector on the Attributes tab doesn't seem to have any effect. It's a little confusing if you aren't aware of the key ability score selector that resides in each class entry. That entry might have to be manually set if you play a class like soldier for which you have a choice of key ability score.

Given that a player could have multiple key ability scores if they have multiple classes, should this be removed? Or perhaps it could be pre-calculated and show all the key ability scores the player has.

ian612:
Removes the old, unneeded key ability score field that was present on actor sheets. These have been managed by class items on PCs, and not used at all by NPCs and drones, for a long time. The field itself did nothing and was somewhat confusing.

While fixing this, updated some localizations related to the key ability score to remove some duplicate fields.

Resolves #223

### DIFF
--- a/static/templates/actors/parts/actor-spellbook.hbs
+++ b/static/templates/actors/parts/actor-spellbook.hbs
@@ -1,10 +1,13 @@
-<div class="inventory-filters spellbook-filters">
-    {{#if isNPC}}
+{{#if isNPC}}
+<div class="npc-ability-field">
     <div class="form-group" data-tooltip='<strong>{{localize "SFRPG.NPCSheet.Header.BaseSpellDC"}}</strong><br>"@attributes.baseSpellDC.value"<br><br>{{localize "SFRPG.NPCSheet.Header.BaseSpellDCTooltip"}}'>
-        <label><h3 class="attribute-name box-title">{{localize "SFRPG.NPCSheet.Header.BaseSpellDC"}}</h4></label>
+        <label><h3 class="box-title">{{localize "SFRPG.NPCSheet.Header.BaseSpellDC"}}</h3></label>
         <input type="number" name="system.attributes.baseSpellDC.base" value="{{system.attributes.baseSpellDC.base}}" placeholder=0/>
     </div>
-    {{else}}
+    <hr>
+</div>
+{{else}}
+<div class="inventory-filters spellbook-filters">
     <div class="form-group spellcasting-ability">
         <label><h4 class="attribute-name box-title">{{localize "SFRPG.SpellBook.SpellCastingAbility"}}</h4></label>
         <select name="system.attributes.spellcasting" data-type="String">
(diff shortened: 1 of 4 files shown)


ian612:
Latest update should fix an issue I found whereby if an actor did not have a class with a key ability score present, skill and save DCs would not calculate correctly.

src/module/rules/actions/item/calculate-save-dc.js
```diff
@@ -19,21 +19,22 @@ export default function(engine) {
...
+
+                    if (itemData.type === "spell") {
+                        dcFormula = "10 + @item.level" + (abilityKey ? ` + @owner.abilities.${abilityKey}.mod` : "");
+                    } else if (itemData.type === "feat") {
+                        dcFormula = "10 + floor(@owner.details.level.value / 2)" + (abilityKey ? ` + @owner.abilities.${abilityKey}.mod` : "");
+                    } else {
+                        dcFormula = "10 + floor(@item.level / 2)" + (abilityKey ? ` + @owner.abilities.${abilityKey}.mod` : "");
+                    }
+
+                    if (actor.type === "npc" || actor.type === "npc2") {
```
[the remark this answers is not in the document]
ian612: Yeah, but the double calculation of dcFormula for NPCs wasn't intentional, so I've updated the PR so this is a bit more logical

danimrath:
This seems good to me but for the life of me for some reason I can't get it to render on my machine. @LebombJames could you try?

static/templates/actors/parts/actor-spellbook.hbs
```diff
@@ -1,11 +1,18 @@
 <div class="inventory-filters spellbook-filters">
+    {{#if isNPC}}
+    <div class="form-group" data-tooltip='<strong>{{localize "SFRPG.NPCSheet.Header.BaseSpellDC"}}</strong><br>"@attributes.baseSpellDC.value"<br><br>{{localize "SFRPG.NPCSheet.Header.BaseSpellDCTooltip"}}'>
+        <label>{{localize "SFRPG.NPCSheet.Header.BaseSpellDC"}}</label>
```
danimrath, line 4: <label><h3 class="attribute-name box-title">{{localize "SFRPG.NPCSheet.Header.BaseSpellDC"}}</h4></label>
        <input type="number" name="system.attributes.baseSpellDC.base" value="{{system.attributes.baseSpellDC.base}}" placeholder=0/>

danimrath: Might as well change the input type to number and looks weird to have the huge input box.

ian612:
I've updated this to move the ability DC to the Features tab, and make the boxes a bit smaller so they don't look as out of place.
image

PR #1604, opened 2025-07-14
Return RollResult objects from Actor/Item roll methods

### LINKED ISSUE #1600: Make roll generation functions in `actor` and `item` return values for use in modules and macros
In SF1E 0.27.3 Changing the actor.rollSkill code from
**return** await this.rollSkillCheck(skillId, skl, options);
to
await this.rollSkillCheck(skillId, skl, options);
Broke a bunch of our homemade macros that used the skill and attack roll results for things like feint and trick attack.
Also broke this https://github.com/nzlbob/sfrpg-galactic-trade

levirak:
3ade256f1e444d972eb24b2b43eb27ad5d9ceb45 broke someone's macros. On reflection, expecting those methods to return values is reasonable, so let's provide a proper API for it.

closes #1600

### DIFF
--- a/src/module/actor/actor.js
+++ b/src/module/actor/actor.js
@@ -555,18 +555,19 @@ export class ActorSFRPG extends Mix(foundry.documents.Actor).with(ActorCondition
      * Prompt the user for input regarding Advantage/Disadvantage and any Situational Bonus
      * @param {string} skillId      The skill id (e.g. "ins")
      * @param {Object} options      Options which configure how the skill check is rolled
+     * @returns {Promise<RollResult?>}
      */
     async rollSkill(skillId, options = {}) {
         const skl = this.system.skills[skillId];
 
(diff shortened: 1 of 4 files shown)


src/module/actor/actor.js
```diff
@@ -555,18 +555,19 @@ export class ActorSFRPG extends Mix(foundry.documents.Actor).with(ActorCondition
      * Prompt the user for input regarding Advantage/Disadvantage and any Situational Bonus
      * @param {string} skillId      The skill id (e.g. "ins")
      * @param {Object} options      Options which configure how the skill check is rolled
+     * @returns {Promise<RollResult?>}
```
LebombJames, line 558: We can add a JSdoc import for `RollResult` across these files, like:
```js
/** @import { RollResult } from <some-path>/dice.mjs */

PR #1622, opened 2025-08-12
Fix renderCombatTracker hook error

Theleruby:
If a non-GM player opens the combat tracker while a combat exists, but before the first round of combat, the render hook currently tries to add event listeners to prev/next combat type buttons that don't exist for that non-GM player. This causes an error to be thrown:

combat tracker error

This PR fixes the issue, preventing the error from appearing.

### DIFF
--- a/src/module/combat/combat.js
+++ b/src/module/combat/combat.js
@@ -927,18 +927,20 @@ Hooks.on('renderCombatTracker', (app, html, data) => {
         // Add buttons for switching combat type
         activeCombat.renderCombatTypeControls(header);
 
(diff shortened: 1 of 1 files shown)


danimrath:
Should we add an isGm check on that section?

ian612:
Yeah I agree, I think an `isGM` check around anything not visible to players would be a better way to get rid of the error.

Theleruby:
Seems fair. I will change it and update my PR

Theleruby:
Updated :)

ian612 (approved):

PR #1626, opened 2025-08-12
Fix turrets not being impacted by critical weapon system damage

Theleruby:
Turrets currently ignore critical weapon system damage when rolling attacks. This PR fixes that issue by setting the modifier for turret weapons to the worst weapon array modifier.

### DIFF
--- a/src/module/rules/actions/actor/starship/calculate-starship-critical-status.js
+++ b/src/module/rules/actions/actor/starship/calculate-starship-critical-status.js
@@ -95,6 +95,15 @@ export default function(engine) {
             systemData.modOther = (key === "powerCore") ? critModsOther[systemData.value] : 0;
         }
 
+        data.attributes.systems.weaponsArrayTurret = {
+            mod: Math.min(
+                data.attributes.systems.weaponsArrayForward.mod,
+                data.attributes.systems.weaponsArrayPort.mod,
+                data.attributes.systems.weaponsArrayStarboard.mod,
+                data.attributes.systems.weaponsArrayAft.mod
+            )
+        };
+
         return fact;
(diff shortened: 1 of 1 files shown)


danimrath:
Is that how it actually works? I thought it just got whatever penalty the facing you chose had?

Theleruby:
CRB pg 321 says:

>When critical damage is scored, the attacking PC should roll on the table below to randomly determine which of the target starship’s key systems is hit; that system gains a critical damage condition (see below), with the effect listed on the table.

>To determine which system is affected, roll d% and consult the table below.

>31–60
>Weapons array 
>Randomly determine one arc containing weapons; condition applies to all gunner actions using weapons in that arc (**a turret counts as being in all arcs**)

We interpreted "a turret counts as being in all arcs" to mean that if e.g. the port arc takes critical damage, the critical condition on the port arc applies to all gunner actions using the turret, since a turret is counted as being in the port arc, even if the turret gets fired in a different direction e.g. the forward arc.

so like:
condition applies to all gunner actions using weapons [mounted] in that arc (a turret counts as being [mounted] in all arcs)
rather than:
condition applies to all gunner actions using weapons [to fire at targets] in that arc

The same interpretation I think would then apply to weapons with the Broad Arc property such as Gyrolaser. You would mount that in e.g. forward arc. Regardless of which direction you fired in, it should be using forward arc penalty because that's where the weapon is mounted.

The wording in the CRB is often a bit vague/unclear though so I can understand why you would interpret it the other way around.

ian612:
This rule definitely needed a clarification or rewording in some errata. I think that this interpretation is correct, though, turrets should take the worst penalty of all the weapon arcs.

danimrath (approved):
Seems fine to me.

PR #1630, opened 2025-08-17
Fix hotbar

Theleruby:
This is a PR for several small hotbar fixes:

- During the upgrade to Foundry v13 the custom hotbar implementation (HotbarSFRPG) was disabled. This pull request makes it compatible with Foundry v13 and re-enables it.
- Fixed the hotbarDrop not working properly
- The tick/cross icon in the corner of macro buttons and the Active/Inactive tooltip label no longer appear for items which have instantaneous duration

hotbar-sfrpg-v13

### DIFF
--- a/src/module/actor/actor.js
+++ b/src/module/actor/actor.js
@@ -297,10 +297,14 @@ export class ActorSFRPG extends Mix(foundry.documents.Actor).with(ActorCondition
 
     /**
      * Delete any of corresponding timedEffect objects of the actor's items.
+     * Also remove linked hotbar instances from item apps before deleting so the hotbar isn't closed.
      */
     async _onDelete(options, userId) {
         for (const item of this.items) {
             game.sfrpg.timedEffects.get(item.uuid)?.delete();
+        }
+        for (const item of this.items) {
+            item.apps?.hotbar?.delete();
         }
 
         return super._onDelete(options, userId);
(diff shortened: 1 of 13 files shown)


src/module/actor/actor.js
```diff
@@ -297,11 +297,15 @@ export class ActorSFRPG extends Mix(foundry.documents.Actor).with(ActorCondition
 
     /**
      * Delete any of corresponding timedEffect objects of the actor's items.
+     * Also remove linked hotbar instances from item apps before deleting so the hotbar isn't closed.
      */
     async _onDelete(options, userId) {
         for (const item of this.items) {
             game.sfrpg.timedEffects.get(item.uuid)?.delete();
         }
+        for (const item of this.items) {
```
danimrath, line 306: Is there a reason you created a second for loop?

ian612: Nope, just didn't register that I was iterating over things twice. I'll combine.

danimrath: Sounds good.

src/module/apps/ui/hotbar.js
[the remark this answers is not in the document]
Theleruby: I tested your change; while the hotbar no longer disappears when deleting the actor, it's now throwing `TypeError: item.apps?.hotbar?.delete is not a function`.
Also, the hotbar is still disappearing upon deleting the items rather than deleting the actor.
I am going to update the PR to fix these issues.

Theleruby:
Updated with localizations, fix for TypeError, and fix for id in hooks to stop deleting items having the same hotbar closing issue

ian612 (approved):
Good catches on the item deletion.

PR #1651, opened 2025-08-24
Improve timed effects

Theleruby:
The timed effects system is quite buggy. This PR tries to address many of the issues that I found while trying to use it.

Bugs/issues addressed:
* Rewrite of handleTimedEffects. The existing implementation is based on comparing the old and new combatants, which relies on the combatants always being iterated over in sequence with nobody being skipped. This causes issues if combatants get skipped over or reordered. It also does not check the status before toggling, which can result in some effects being incorrectly toggled (this is most visible if you have to rewind). The rewrite instead attempts to work out if an effect should be active and compares this to the current state, toggling if it's no longer correct. I found this to be a lot more reliable.
* When starting a combat, the world time is advanced the same way that it would be when starting a new round. This fixes some issues with effects being incorrectly toggled or getting the wrong duration if you toggle them outside of combat and then start a new combat.
* Manually deactivating feats now resets their timed effects. The regular timed effect objects were already working this way so this allows more consistent behaviour.
* When activating a feat, the actor whose turn it was activated on is now stored. This allows it to be deactivated at the correct point if you have to wind the combat backwards for some reason.
* The start of the first round of combat now calls the method to handle timed effects as this was missing.
* Fixed updateWorldTime not checking if the player is the active GM before toggling or poking timed effects.
* Added a new duration to the feats called "turn" which causes the feat activation to last until the end of the turn it's activated on.

I would appreciate some help testing this before this gets merged as I suspect there are probably edge cases which still need addressing.

### DIFF
--- a/static/lang/de.json
+++ b/static/lang/de.json
@@ -993,6 +993,7 @@
             }
         },
         "DurationTypesInstantaneous": "Augenblicklich",
+        "DurationTypesTurn": "Turn",
         "Effect": {
             "DetailsEnabledLabel": "Aktiviert",
             "DetailsHeader": "Effektdetails",
--- a/static/lang/es.json
+++ b/static/lang/es.json
@@ -989,6 +989,7 @@
             }
         },
         "DurationTypesInstantaneous": "Instantáneo",
+        "DurationTypesTurn": "Turn",
         "Effect": {
             "DetailsEnabledLabel": "Habilitado",
             "DetailsHeader": "Detalles del efecto",
--- a/static/lang/fr.json
+++ b/static/lang/fr.json
@@ -991,6 +991,7 @@
             }
         },
         "DurationTypesInstantaneous": "Instantané",
+        "DurationTypesTurn": "Turn",
         "Effect": {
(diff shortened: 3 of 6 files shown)


LebombJames:
Not sure how I feel about advancing 30 mins at the end of combat. Perhaps an "expires on combat end" toggle on effects would solve the problem more cleanly

Theleruby:
I accept the time advancement was a bit of a bodge. It was hard trying to make a decision on how to solve clearing out the effects... I think whether you want them to end or not is going to be personal preference, and either way we should avoid clearing out longer duration effects e.g. those lasting for many hours.

I think if you wanted that toggle, it would be quite a lot of work to add to all the effects and feats, as well as making a decision on which ones it should be applicable for. Maybe that can be automated somehow though.

Theleruby:
I think that should be pretty straight forward yes.

ian612:
Cool, I think that (and a setting to enable/disable the behaviour) should be a good compromise, then

ian612:
Cool sounds good, thanks for looking at that. No worries at all on the timing, take as much as you need 🤙

PR #1658, opened 2025-08-28
feat: Add starship modifier support

CharlesHunter:
Localization text generated from AI.
Tested and works on a local dev build of FVTT 13.347

image

image


LebombJames:
Code looks good, but please leave non-English localisations as using the English text as a placeholder rather than translating it yourself (use the `copyLocalization` npm script)

ian612:
I've tested this as well, and everything looks good except the Starship Piloting Skill Modifier; it works fine with PC starships but doesn't seem to get calculated in to the values for ships with NPC crews.

ian612 (changes requested):
I've tested this as well, and everything looks good except the Starship Piloting Skill Modifier; it works fine with PC starships but doesn't seem to get calculated in to the values for ships with NPC crews.

ian612:
Looking at this in a bit more detail, it looks like calculating the piloting modifier for NPC rolls would require a structural change to how NPC Crew data is stored if we want to do it cleanly. I've added a TODO comment noting what would need to be done to get that to work, as it will likely have to wait until after DataModels are implemented and easier migration can be done for the NPC Crew data structure.

PR #1659, opened 2025-09-06
Fix condition application issues

### LINKED ISSUE #1002: Conditions switch on the token HUD don't work when translated with Babele
**Describe the bug**
modules : Babele + sfrpg-fr-translation (https://github.com/aldarion-jdr/sfrpg-fr-translation)
Conditions Compedium is translated, Actor's sheet is correct and when I click a conditions on the Actor's sheet it adds the conditions to the actor.
Problem is when I use the HUD, the condition icon is added to the Token, but not in the actor's sheet.

**To Reproduce**
1. Right click on Token
2. Click on one condition
3. Condition is not synchronized with related Token's actor sheet

**Expected behavior**
Condition selected on the HUD should be synchronized with the Tokens's actor sheet

**Screenshots**
image

 - OS: Windows 10
 - Browser: all+Foundry app
 - Foundry Version: 10.291
 - SFRPG Version: 0.22.1

**Additional context**
None other

ian612:
This attempts to resolve a race condition by adding an additional await in condition application code. In addition, condition application code is cleaned up a bit.

A `nameSlug` field is added to the condition items so they can be referenced by this instead of their name, which should allow for more robust translation of item names in the compendium.

Resolves #1002

### DIFF
--- a/src/module/actor/mixins/actor-conditions.js
+++ b/src/module/actor/mixins/actor-conditions.js
@@ -27,7 +27,7 @@ export const ActorConditionsMixin = (superclass) => class extends superclass {
             return undefined;
         }
 
-        const conditionItem = this.items.find(item => item.name.toLowerCase() === conditionName);
+        const conditionItem = this.items.find(item => (item.system.nameSlug === conditionName) || (item.name.toLowerCase() === conditionName)); // TODO: The 'or' is in place for backward compatibility. This should only use nameSlug once DataModels are implemented and migrations can be done easily.
         return conditionItem ?? null;
     }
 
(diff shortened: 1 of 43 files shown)


ian612:
Change `nameSlug` to `slug` to better match up with changes in #1602

src/module/actor/mixins/actor-conditions.js
```diff
@@ -12,7 +12,7 @@ export const ActorConditionsMixin = (superclass) => class extends superclass {
         }
 
         const conditionItem = this.getCondition(conditionName);
-        return (conditionItem !== undefined);
+        return conditionItem ?? null;
```
LebombJames, line 15: Let's keep this returning a boolean, unless there's a particular reason for this change

ian612: I think I changed it because I had `getCondition()` return `null` if it didn't find anything instead of undefined. I agree returning `true` or `false` here makes more sense, though, so I've made the recommended change.

PR #1662, opened 2025-09-12
Automatically select combat type when GM adds actors to combat

ian612:
Changes the combat type automatically based on the type of actor most recently added to a combat by the GM.


Theleruby:
I think maybe it's worth putting a safety check in so that this doesn't execute if the combat is already started

ian612:
Ah true, very good point. I'll add that in

Issue #1414, opened 2024-12-22
Consumable Items always make damage rolls or saving throws when used


ian612:
**Describe the bug**
When a consumable item is used, it sends a message to the chat. The code seems as though it always makes a damage-type roll regardless of what the action type is set to, except for saving throws.

This seems to be related to a line in `rollConsumable(options = {})` in `item.js`, where an if statement lumps any action type that is not a saving throw together, and tries to make a damage roll with it.

**To Reproduce**
Create a consumable item, set action type to Melee Attack.

**Expected behavior**
I expect an attack roll to be made with the appropriate bonuses.

**Actual behavior**
The system attempts to make a damage roll, which throws an error message in Foundry.

**Technical details (please complete the following information):**
 - OS: Windows 11
 - Browser: Chrome
 - Foundry Version: 12.331
 - SFRPG Version: v27.0 (dev build)

PR #1666, opened 2025-09-26
Allow expenditure of ammo/charges by items from actor sheet

### LINKED ISSUE #580: Armour Upgrades can be given ammunition, but can't expend it
Some armour upgrade items expend charges at a specific rate. Some of these will be a passive cost for being powered on, but others are active costs that require an action (or are used as part of another action) to activate, such as the Jump Jets. However, as no option appears to activate the upgrade, either while it's mounted in an armour upgrade slot or while it's just in the inventory, charges can't be expended.

**Steps to reproduce:**

1. Move Jump Jets from the compendium to a character's sheet
2. Note that the Jump Jets have no activation button, yet their capacity and usage are set to 20 and 2/Action respectively
3. Set the ammunition type of the upgrade to Charges
4. Add a storage slot to the upgrade, set to accept ammunition
5. Add a battery from the compendium to the character's sheet
6. Reload the Jump Jets
7. Note that the Jump Jets now contain a battery and are fully charged, but have no activation button
8. Drag the upgrade onto the hotbar to create a macro
9. Use the macro to activate the Jump Jets
10. Note that this posts the Jump Jets' description to the chat rather than activating them

ian612:
Resolves #580

### DIFF
--- a/src/module/item/item.js
+++ b/src/module/item/item.js
@@ -1669,6 +1669,13 @@ export class ItemSFRPG extends Mix(foundry.documents.Item).with(ItemActivationMi
     async useItem(options = {}) {
         const itemData = this.system;
         const overrideUsage = !!options?.event?.shiftKey;
+        const overrideAddCapacity = options?.event?.altKey;
+
+        // Instead of activating the item, add charges to it
+        if (overrideAddCapacity) {
+            this.increaseCapacity(this.type === "consumable" ? 1 : itemData.usage.value);
+            return;
+        }
 
         let currentCapacity = 0;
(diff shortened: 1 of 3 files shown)


ian612:
Yep this should cover that, you can click the button on the utilisation column to deduct charges.

image

ian612:
I've just spent some time streamlining the `rollConsumable()` functionality into this new code to unify it with the newer `useItem()` since both functions essentially do the same thing. I'm now realizing that I'd like to solve #1414 here since I'm poking around at that code anyway, but I'll also give your suggestions a go, as I think they're worthwhile additions.

PR #1691, opened 2025-11-13
Fix a deepClone/dataModels bug where the system overwrites all items

ian612:
This fixes an issue that was causing all items to be overwritten with the data of an item when an attack or damage was rolled with that item.

Original sheet:
image

Sheet after rolling an attack and advancing the combat turn:
image

It was caused by the implementation of a Data Model for actors causing the `deepClone()`-ing of their system data to stop functioning correctly. A pass was made to remove these instances previously, but two were missed. The solution was to remove these two instances. Fortunately, the functionality these lines originally performed looks to have been made obsolete a while ago, so they could be removed easily without affecting functionality (as far as I can tell).

### DIFF
--- a/src/module/item/item.js
+++ b/src/module/item/item.js
@@ -944,16 +944,13 @@ export class ItemSFRPG extends Mix(foundry.documents.Item).with(ItemActivationMi
             rollOptions.actionTargetSource = SFRPG.actionTargets;
         }
 
-        // Define Roll Data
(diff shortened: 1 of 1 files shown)


ian612:
In making this change, it by consequence removes the ability to directly reference `@mod` within damage formulas. I can only see 8 instances in the compendiums where this is used, so I'm assuming it's safe to remove this and just update the compendium aliens with the correct values.

image

PR #1701, opened 2025-11-25
Equipment Descriptions Update

TheWaywardAldori:
Added Descriptions and Updated Hp to their Maximum totals so they are not considered broken for all Acid Cannons

### DIFF
--- a/src/items/equipment/acid_cannon_dinergate.json
+++ b/src/items/equipment/acid_cannon_dinergate.json
@@ -33,8 +33,8 @@
         "value": ""
       },
       "hp": {
-        "max": "",
-        "value": null
+        "max": "51",
+        "value": 51
       },
       "size": "medium",
       "sturdy": true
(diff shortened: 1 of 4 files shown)


TheWaywardAldori (commented):
Forgot to cook before submitting these

src/items/equipment/acid_cannon_dinergate.json
```diff
@@ -34,7 +34,7 @@
       },
       "hp": {
         "max": "",
-        "value": 6
+        "value": null
```
danimrath: shouldn't this be 51?

src/items/equipment/acid_cannon_ergatoid.json
```diff
@@ -34,7 +34,7 @@
       },
       "hp": {
         "max": "",
-        "value": 6
+        "value": 96
```
danimrath: Isn't this 66? level * 3 + 15?

src/items/equipment/acid_cannon_dinergate.json
ian612: Yeah it should be, I've updated it

ian612 (approved):

PR #1710, opened 2025-12-04
Fix remaining DR Issues

### LINKED ISSUE #548: Specific damage reduction
The enhanced feat is not correctly written in the foundry system. It should be for kinetic damage either piercing, slashing or bludgeoning (not all of them)

Furthermore, I can't find how to implement slashing, piercing, or bludgeoning DR

### LINKED ISSUE #563: Missing Damage Reduction and Weapon Properties for Magic, Wound, Chaotic, Good, Evil, Lawful and Wound.
In regards to Weapons (Under Special materials) and the token properties for Damage Reduction.
It appears the following are missing:
-	Magic
-	Good
-	Evil
-	Chaotic
-	Lawful
-	Wound (example Cybernetic Golem AA2 Pg. 66) 
I know, Wound and other weapon properties aren’t a common weapon reduction. Which you could lump under Custom. But Magic, Chaotic, Good, Evil and Lawful are common enough that it necessitate there own category. 

Also, easier said than done. But may need to require an **and** statement for damage reduction. For example: Endbringer Devil (Dhalochar) AA1 Pg 34. It has a DR: 15/ Good **and** Magic.

### LINKED ISSUE #679: Shields unable to receive Basic Melee bonuses when attacking.
I was recently able to update all the shields that allow attacks as unarmed attacks to do basic damage. However, since they do attacks as unarmed attacks. Shield items will need, in some means, the ability to receive bonuses from Basic Melee attacks/damage modifiers. Such as from Specialization.

### LINKED ISSUE #1092: Remaining DR issues
- [ ] Magic, Bludgeoning, Piercing, Slashing, the different alignments, missing from Damage Reduction Modifiers list.
- [ ] Magic, the different alignments, check boxes missing from Weapon details
- [ ] Spell damage to all be treated as Magic for DR
- [ ] "Or" option added to Damage Reduction. Current system only allows "And" for DR exceptions, but "Or" is also possible.
- [ ] DR in attributes tab only displays one DR exception currently.

ian612:
This PR resolves the system's remaining DR issues listed in #1092.

It adds more damage reduction options, adds descriptors to items with actions and checks these when applying damage, and automatically calculates whether an item's damage is magical in nature.

Remaining Tasks:

- [x] Allow for DR to be bypassed by any of the selected damage types rather than only by all of the damage types
- [x] Display at least two DR entries on character sheet

Resolves #1092
Resolves #679
Resolves #563 
Resolves #548


src/module/item/item.js
```diff
@@ -96,6 +96,21 @@ export class ItemSFRPG extends Mix(foundry.documents.Item).with(ItemActivationMi
...
 
+    /**
+     * Does the Item's damage qualify as magic?
+     * @type {boolean}
+     */
+    get hasMagicDamage() {
+        if (!this.hasDamage) return false;
+        if (['magic', 'hybrid', 'spell'].includes(this.type)) return true;
+        if (['msak', 'rsak'].includes(this.system.actionType ?? null)) return true;
+        const containedItems = this.contents ?? [];
```
danimrath: Need to add this.system.properties.hybrid.value check

src/items/class-features/armored_form_(ex).json
```diff
@@ -55,6 +55,7 @@
         "name": "Armored Form (Ex) - Bludgeoning",
         "type": "untyped",
```
danimrath: these should probably use like the same type and not stack.

src/items/alien-archives/spec-ops_veteran.json
```diff
@@ -2907,25 +2907,7 @@
       "source": "IS pg. 165",
       "subtype": "Shapechanger"
     },
-    "modifiers": [
```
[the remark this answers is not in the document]
ian612: This one has the same DR on an item and directly on the actor, I just removed one since the one on the item should be sufficient and they don't stack anyway

danimrath (approved):
I think this is pretty good.

PR #1717, opened 2025-12-17
Attack roll display hit

### LINKED ISSUE #1475: Show an indication of whether an attack roll hits or not on chat cards
Every Foundry System I've used has attack roll automation - that is to say, on an attack, it'll say if it succeeds or not, sometimes by how much. I can't find it in this system - when attacking a target, it just says the calculated result of the roll.

If there's any chance to add a feature to show if the roll was a success/failure/crit, I think it would go a long way for the already-impressive program

ian612:
This PR adds attack roll to-hit display to chat cards when a token is targeted by an attack roll. The process for doing this involves:

- [x] Add first available `game.user.targets` to the roll context object as `target`
- [x] Add a new `rollType` property to be chosen at time of roll initiation to be passed down the chain to all relevant roll functions/chat cards
- [x] Display the selected target on the roll dialog if making an attack roll
- [x] Using the roll's `actionTarget`, evaluate success/failure of the roll being made for cha
- [x] Display success/failure on chat card
- [x] Add support for starships
- [x] Same as above, but on custom chat cards

The aim is to have this work for all actor types, including starships and vehicles.

Resolves #1475


ian612:
I think functionally this is now done; it will need prettifying but I think that can wait until later on, after the chat cards have been redone.

danimrath:
Could we add in for GM only the AC that was used in the calculation to the chat card? That would probably be a good sanity check.

ian612:
On some playtesting, this should also show success/fail on chat card buttons that trigger a skill check or saving throw.

ian612:
> On some playtesting, this should also show success/fail on chat card buttons that trigger a skill check or saving throw.

This should now work on saves and skill checks rolled with chat card buttons after a couple additions and a little bit of re-organization. The last thing I want to do with this is set up automatic showing/hiding of the check DC / AC of the creature based on ownership and/or GM status. I think this should be possible via attaching some render event listeners in the item chat mixin.

ian612:
Couldn't find any issues on final testing, so going to merge this now

PR #1720, opened 2025-12-23
Implement @Condition enricher

LebombJames:
- Refactor custom enrichers to bring them inline with V12/V13 changes
- Also implement support for event listeners other than click

`@Condition` mandated the implementation of a cache for condition items. This should probably be used to add conditions to actors, instead of always fetching from the compendiums


src/module/system/enrichers/base.js
```diff
@@ -189,45 +180,34 @@ export default class BaseEnricher {
...
-    static addListeners() {
-        const body = $("body");
-        body.on("click", `i.repost`, this.repostListener);
-        for (const [action, cls] of Object.entries(CONFIG.SFRPG.enricherTypes)) {
-            if (cls.hasListener) body.on("click", `a[data-action="${action}"]`, cls.listener);
-        }
-    }
+export function getDatasetfromEvent(event) {
+    const anchorEl = event.currentTarget.children[0];
+    return anchorEl?.dataset || null;
```
ian612: This doesn't seem to be getting any `duration` or `unit` fields from the dataset, so they aren't being set

src/module/token/token-ruler.js
```diff
@@ -119,7 +119,7 @@ export default class SFRPGTokenRuler extends foundry.canvas.placeables.tokens.To
                     return false;
                 }
             }
-        }, { performDeletions: true });
+        }, { applyOperators: true });
```
danimrath, line 122: What does this change do?

PR #1725, opened 2026-01-11
Issue 1567 Track expanded state on documents and items

### LINKED ISSUE #1567: Feats expanded in sheet close when combat turn changes
When a feat/feature is expanded in sheet and the turn in combat switches whatsoever, the feat minimizes immediately.

To Reproduce:
1. Start an encounter with any two actors (tested with PC and NPC sheets)
2. Open any sheet and click on a feat/feature/entry with text in its description.
3. With the feature's description displayed underneath its name, progress the turn tracker.
4. Feature's expanded description will close immediately upon step 3.

The sudden closing of feats between turns is disorienting, as the feats are expected to be independent from initiative turn tracking.

Image
_Screencap of Foundry's built-in turn tracker and a test character sheet. Ambuscade feat is expanded, but this happens with all tested entries._

Image
_Same screencap, but the turn has progressed to the Goblin's. The feat minimizes as soon as this happens._

 - OS: Windows 10
 - Browser: Firefox, Chrome, Internet Explorer, and the Foundry Client itself
 - Foundry Version: 12.343
 - SFRPG Version: 0.27.3

danimrath:
These changes make it so whenever a sheet re-renders even after being closed it persists the expanded state of the documents/items.

Resolves #1567 

This Change also includes adding a new setting to have all item summaries close when a sheet is closed or to remember the closed/open state.

To add to that a new keybinding can be found in the controls section that defaults to CTRL+SHIFT+C that closes all open summaries on an active actor sheet.

### DIFF
--- a/src/module/actor/sheet/base.js
+++ b/src/module/actor/sheet/base.js
@@ -85,6 +85,10 @@ export class ActorSheetSFRPG extends foundry.appv1.sheets.ActorSheet {
         data.labels = this.actor.labels || {};
         data.filters = this._filters;
 
+        for (const item of data.items) {
+            item.expanded = item.getFlag('sfrpg', 'expanded') || false;
+        }
+
         if (!data.system?.details?.biography?.fullBodyImage) {
             this.actor.system = foundry.utils.mergeObject(this.actor.system, {
                 details: {
@@ -157,6 +161,25 @@ export class ActorSheetSFRPG extends foundry.appv1.sheets.ActorSheet {
 
         this._prepareItems(data);
 
+        // Prepare summary HTML for expanded items
+        for (const item of data.items) {
+            if (item.expanded) {
+                const chatData = await item.getChatData();
+                const desiredDescription = chatData.description.short || chatData.description.value;
+                const div = $(`<div class="item-summary">${desiredDescription}</div>`);
+                const props = $(`<div class="item-properties"></div>`);
+                chatData.properties.forEach(p => {
+                    let tooltipValue = p.tooltip || p.title || "";
+                    if (tooltipValue) tooltipValue = `data-tooltip="${tooltipValue}"`;
+                    props.append(
+                        `<span class="tag" ${tooltipValue}><strong>${p.title ? p.title + ":" : ""} </strong>${p.name}</span>`
+                    );
+                });
+                div.append(props);
+                item.summaryHTML = div[0].outerHTML;
+            }
+        }
+
         // Enrich text editors. The below are used for character, drone and npc(2). Other types use editors defined in their class.
         const secrets = this.actor.isOwner;
         data.enrichedBiography = await foundry.applications.ux.TextEditor.enrichHTML(this.actor.system.details.biography.value, {
@@ -170,6 +193,7 @@ export class ActorSheetSFRPG extends foundry.appv1.sheets.ActorSheet {
             secrets
         });
 
+        this.data = data;
         return data;
     }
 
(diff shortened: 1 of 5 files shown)


ian612:
I've noticed that the summary div looks to be different between the version that's generated by `_onItemSummary()` when first clicking on the item (top image) and the version that's generated in `activateListeners()` that displays when the div is "persisted" through a combat round change, e.g. (bottom image). Something funny also seems to be going on with the calculated attack & damage bonuses, though I'm not sure if that's related to this PR or something else.

EDIT: The whole issue seems to be related to something else, I'm going to try to fix that a push a hotfix for it for now and then come back to this to see if it's working correctly afterward

image

image

src/module/apps/item-collection-sheet.js
```diff
@@ -130,12 +132,26 @@ export class ItemCollectionSheet extends DocumentSheet {
             }
         }
 
+        // Prepare summary HTML for expanded items
```
ian612: I actually don't think we need any changes on this sheet; the sheet re-renders (collapsing summaries) only happen on actors who are in a combat, while item collections should never be in combat.

I gave this a test and the previous version of the item collection sheet seems to work fine, persisting summaries without issue.

src/module/actor/sheet/base.js
```diff
@@ -203,6 +235,17 @@ export class ActorSheetSFRPG extends foundry.appv1.sheets.ActorSheet {
     async activateListeners(html) {
         super.activateListeners(html);
 
+        // Add summary divs for initially expanded items
+        html.find('.item').each((i, el) => {
+            const li = $(el);
+            const itemId = li.data('item-id');
+            const item = this.data.items.find(x => x.id === itemId);
```
ian612: Can this just be `this.actor.items.find(...);` ?

src/module/actor/sheet/base.js
```diff
@@ -1050,16 +1096,17 @@ export class ActorSheetSFRPG extends foundry.appv1.sheets.ActorSheet {
...
             div.append(props);
             li.append(div.hide());
 
             div.slideDown(200, function() { /* noop */ });
+            li.addClass('expanded');
+            item.expanded = true;
         }
-        li.toggleClass('expanded');
 
+        await item.setFlag('sfrpg', 'expanded', item.expanded);
```
ian612: I'm still getting the jumpiness when expanding/shrinking the items. I think this line is running async and triggering a re-render of the sheet. I had something working in a PR I posted on your branch via putting this into the callback of the slideUp/slideDown function, which I think should fix it

src/module/actor/sheet/base.js
```diff
@@ -170,6 +201,7 @@ export class ActorSheetSFRPG extends foundry.appv1.sheets.ActorSheet {
             secrets
         });
 
+        this.data = data;
```
ian612: I don't think this is needed, it's only used once, later on line 242, where it can be replaced by `this.actor.items`

src/module/actor/sheet/base.js
```diff
@@ -85,6 +93,10 @@ export class ActorSheetSFRPG extends foundry.appv1.sheets.ActorSheet {
         data.labels = this.actor.labels || {};
         data.filters = this._filters;
 
+        for (const item of data.items) {
+            item.expanded = item.getFlag('sfrpg', 'expanded') || false;
```
ian612: This could be moved down to be put into the same loop as on line 165 so it doesn't have to run twice

src/module/actor/sheet/base.js
ian612: Can this just be `this.actor.items.find(...);` ?

src/module/actor/sheet/base.js
danimrath: Thanks, wasn't getting a notification about the PR.

ian612 (approved):
Looks good to me

PR #1731, opened 2026-01-26
Convert select client settings to user settings where it makes sense

ian612:
Since user settings are now available, makes sense to have some of the system settings convert to that type.


danimrath (approved):
I see no problem with these.

PR #1733, opened 2026-01-26
Remove system's compendium art mapping

### LINKED ISSUE #1695: Deprecate system-specific art mapping in favor of Foundry native mapping
Foundry introduced compendium art mapping as core functionality at some point in the past, and as such it makes sense to defer to their implementation.

There are two paths forward:
1. We simply remove all our compendium art mapping code
2. We set up a method that "forwards" any art mappings detected by the system in the old format to Foundry's core implementation

**Additional context**
It would also be wise, once a method is chosen, to let Metamorphic (premium content) and Daedril (pdf importer module author) know about the decision.

ian612:
This PR remove's the system-specific compendium art mapping functionality to defer to Foundry's native implementation.

Included is also a macro to assist users converting their old-style art file's in the sfrpg format to Foundry's format. Once this is implemented, I'll also update the [wiki page here](https://github.com/foundryvtt-starfinder/foundryvtt-starfinder/wiki/Compendium-Image-Mapping).


src/module/item/item.js
```diff
@@ -18,13 +18,6 @@ import { ItemChatMixin } from "./mixins/item-chat.js";
 export class ItemSFRPG extends Mix(foundry.documents.Item).with(ItemActivationMixin, ItemCapacityMixin, ItemChatMixin) {
 
     constructor(data, context = {}) {
```
ian612: If there's nothing in them, do these constructors need to be here?

danimrath: I don't think so.

PR #1738, opened 2026-02-02
Add ChatMessageSFRPG class and data model

ian612:
This adds a `ChatMessageSFRPG` class and corresponding data model to store the constructing system data for chat messages. This will allow for messages to be more dynamic and support more functionality while staying organized (i.e. not packing all data into the message's flags or html data fields).


src/module/data/message/base-message.mjs
```diff
@@ -0,0 +1,84 @@
...
+        // TODO: We may want to change this to an array of strings at some point
+        schema.rollBreakdown = new fields.StringField({
+            initial: "",
+            blank: true
+        });
+
+        // TODO: Eventually, add more detail to this
+        schema.rollCriteria = new fields.ObjectField();
+
+        schema.tags = new fields.ArrayField(
```
LebombJames: Could this be a `TypedObjectField` with the key being the tag? `ArrayField`s scare me

ian612: It could be, but I've been using a handlebars `#each` helper to display them on the chat cards. Changing to a TypedObjectField would mean I'd have to generate the tags as an object then parse into an array and store both on the chat message object rather than just generating straight to an array, so it adds an extra step

LebombJames: If i'm not mistaken, using `#each` on an object iterates through it as if you were using `Object.entries`, so you get the key and the object itself as two iteration variables. 

It's also worth mentioning that `renderHTML` can be used as an equivalent to a sheet's `getData`/`_prepareContext` where you format the document's data into the format you want for the template, the document itself doesn't need to the be direct input to the template

ian612: Oh interesting, I didn't realize that `#each` worked like that. In that case yeah, shouldn't be an issue to make it an object.

Regarding the second point, that's what I initially assumed but when I'd try to add fields to the message object they'd only show up during handlebars rendering if I added them to the system data model explicitly. I'll give it another try though, as I think it makes more sense to not store any of the rendering data on the document if possible.

ian612: Ah I see the issue; additional data fields have to be passed in explicitly to `super.renderHTML({canDelete, canClose, ...rest})` in `...rest` for them to be available, as is sanitizes the object with the document data model via a `toObject()` call before packing in the extra data fields and rendering. I should be able to make that change and trim down the data models significantly, in that case.

ian612: Made the change, looks to be working as expected

ian612:
Still need to do a few things before this is functional and complete:

- [x] Add roll breakdowns back to chat cards
- [x] Add damage type display back to damage rolls
- [ ] Add attack action type (`melee, ranged, weapon, spell`), range/range increment, area, ammunition, and usage to attack d20 roll card, and any other cards where it makes sense
- [ ] Add item toChat chat card along with displayed buttons to make skill rolls, attacks, saves, etc.
- [ ] Add "apply damage" buttons to chat card for targeted enemies
- [ ] Add item use chat card
- [ ] Add spellcasting chat card
- [ ] Add reload chat card
- [ ] Add initiative and combat phases/rounds chat cards
- [ ] Add css for formatting cards nicely

ian612:
I was debating rewriting some parts of the roll system as part of this but I think roll nodes/trees/modifiers are too deeply embedded in the codeto pull them out without significant rework, which probably needs to go in another PR.

src/module/data/message/base-message.mjs
```diff
@@ -6,6 +6,11 @@ export default class SFRPGMessageBase extends SFRPGDocumentBase {
     static defineSchema() {
         const schema = super.defineSchema();
 
+        // TODO: Eventually, add more detail to this
+        schema.actor = new fields.ObjectField({ // TODO: Detail this if possible
```
LebombJames, line 10: What's expected to go in this and `schema.item`? Is it (UU)IDs or full actor data?

ian612: For now I'm just packing a full actor/item object in there so I can reference data on the chat cards easily where needed. My plan is to eventually convert this to UUIDs and then store only the exact data fields needed from the actor/item documents on the chat message separately

LebombJames: Wouldn't it be easier to store *only* the UUID, then fetch the actor/item and refer to the fields on it as and when needed? That'll keep the message data model as small as possible

ian612: For most stuff on actors/items that's static (name, icon, etc.), definitely possible. Some things like ammo count I'd want to store statically, though, as I don't want the html rendering process to update it each time the item is updated.

ian612: Actually also would there be a concern for NPC actors whose tokens are deleted mid combat (i.e. if defeated and the GM deletes instead of marking them dead)? Thinking that might cause issues with getting the speaker name and image since the new rendering method is more dynamic than the old one

PR #1739, opened 2026-02-09
Claude/add starfinder mech support

worldsofwondergames:


src/items/mech-components/acid_dart_rifle.json
```diff
@@ -0,0 +1,44 @@
+{
+  "_id": "mechWpn16aciddart",
```
ian612: These need to all be standard Foundry 16-character UUID strings that are unique to each item

src/items/mech-components/acid_dart_rifle.json
```diff
@@ -0,0 +1,44 @@
...
+  "name": "Acid Dart Rifle",
+  "type": "mechWeapon",
+  "folder": "mechFldrWpnRngd1",
```
ian612: Likewise, this needs to be in the format of a standard 16-character Foundry UUID unique to the folder

src/items/mech-components/_folders.json
```diff
@@ -0,0 +1,102 @@
+[
+  {
+    "_id": "mechFldPwrCore01",
```
ian612: These `_id` strings need to be in the format of a standard 16-character Foundry UUID unique to each folder

ian612: Folders are missing `description` fields and the `_stats` object. If the compendium exists within Foundry and you run `npm run unpack`, it should add these fields correctly for you

ian612 (changes requested):
Hello, thanks for submitting this feature. As a first step toward reviewing it, it looks like there are a bunch of extraneous files in a `templates` folder and in a `lang` folder at the base level.
Lastly, could you please remove these as well as the `packs` folder and the extraneous js and map files from the base level of your commit?

Once that stuff's cleaned up, we'll proceed to review further; thanks!

src/items/mech-components/_folders.json
worldsofwondergames: I was actually intending to merge it to my local fork's main branch, but I went ahead and made the fixes as recommended. Apologies for the mix-up.

danimrath:
Just to clarify Ian's comment. You appear to have accidentally added a full copy of the templates folder into the base files. You will need to remove these.
image

ian612:
Also, I can see that you're still working on a few things so just going to convert this to draft for now. If you do want us to review it and merge into the system, just change it back to ready for review whenever you're finished with it.

PR #1740, opened 2026-02-11
Avoid updating actor flags if in a compendium.

danimrath:
When opening an actor, interacting with item summaries or closing an actor we should not edit the flags for persisting open/close state if the actor is in a compendium.


ian612 (approved):
Seems good to me on testing

PR #1750, opened 2026-05-27
Fix drone ownership not saving

### LINKED ISSUE #1743: Allow Drone "Owner" field to have proper association with an Actor
Seems a feature that wasn't ever fully implemented.  Drone sheet shows "Owner" but does nothing.  

Believe this should help auto level the drone based upon appropriate classes by owner actor (technomancer or mechanic)...

ian612:
Fixes drone ownership not saving by adding a UUID field for the drone's owner. Allows an actor to be dropped on the drone's sheet to set ownership, with click-to-open owner sheet functionality, and right-click to clear ownership.

### DIFF
--- a/src/less/actors.less
+++ b/src/less/actors.less
@@ -40,7 +40,11 @@
 
             // For Drone owner name clickability
             li.drone-owner {
+                font-weight: bold;
                 span.drone-owner-name {
+                    font-family: "Exo 2";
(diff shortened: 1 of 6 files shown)


danimrath (approved):
Looks good to me. I would just add some color to the owner name so it's obvious it's interactable.

PR #1752, opened 2026-06-02
Use Item piles for all canvas drops

danimrath:
Testing V14 updates and noticed some issues when dropping items onto the canvas or item piles.


ian612 (approved):
Looks good to me!

PR #1753, opened 2026-06-02
Enricher fixes

ian612:
Fixes an untracked issue where dynamically calculated links were getting stuck as static values after an item was sent to chat.

Caused by fixes made when `deepClone()` was no longer allowed to be used on the `system` property directly.


danimrath (approved):
This looks okay to me.

PR #1762, opened 2026-06-15
Automated Actor Resource Hotbar Macro

ian612:
This PR adds the ability to drag an actor resource item to the hotbar for monitoring and management. Click to increase, Ctrl+click to decrease.

image

### DIFF
--- a/src/module/item/item.js
+++ b/src/module/item/item.js
@@ -1143,18 +1143,6 @@ export class ItemSFRPG extends Mix(foundry.documents.Item).with(ItemActivationMi
     }
 
     /**
-     * Adjust an actor resource value up or down by a value
-     *
-     * @param {Number}      adjustValue The value to adjust the resource by
-     * @returns {Boolean}               True if adjusted, false if wrong item type
-     */
-    adjustResourceValue(adjustValue) {
-        if (this.type !== "actorResource") return false;
-        const newValue = this.system.base + adjustValue;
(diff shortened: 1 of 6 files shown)


src/module/item/item.js
```diff
@@ -1142,6 +1142,18 @@ export class ItemSFRPG extends Mix(foundry.documents.Item).with(ItemActivationMi
         this.consumeCapacity(this._calculateAmmoUsageWithModifiers(usage.value));
     }
 
+    /**
+     * Adjust an actor resource value up or down by a value
+     *
+     * @param {Number}      adjustValue The value to adjust the resource by
+     * @returns {Boolean}               True if adjusted, false if wrong item type
+     */
+    adjustResourceValue(adjustValue) {
```
LebombJames: You could put this on the ActorResource data model, and then refer to it with `item.system.adjustResourceValue`. Makes more sense imo in terms of organisation

ian612: Makes sense. How would you get a `this.update({})` function to work from within the datamodel, when the document and its methods aren't accessible from within it?

LebombJames: `this.parent` on a data model refers to the document. So `this.parent.update`

ian612: Got it, I agree this makes a bit more sense. I've made the change and confirmed it's working

PR #1768, opened 2026-07-01
Temporary Effects returned in appliedEffects

danimrath:
When foundry pulls back the effects it no longer uses temporaryEffects it uses appliedEffects now and checks for the showIcon condition.

### DIFF
--- a/src/module/actor/actor.js
+++ b/src/module/actor/actor.js
@@ -38,7 +38,7 @@ export class ActorSFRPG extends Mix(foundry.documents.Actor).with(ActorCondition
         super(data, context);
     }
 
-    // appliedEffects effects can be displayed on the token, so hijack it and include effects
+    // appliedEffects effects can be displayed on the token, so hijack it and include shown effects
     get appliedEffects() {
         const fromEffects = this.items
             .filter((e) => e.type === "effect" && e.system.showOnToken && e.system.enabled)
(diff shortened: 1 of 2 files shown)


src/module/actor/actor.js
```diff
@@ -38,6 +38,15 @@ export class ActorSFRPG extends Mix(foundry.documents.Actor).with(ActorCondition
...
+    get appliedEffects() {
+        const fromEffects = this.items
+            .filter((e) => e.type === "effect" && e.system.showOnToken && e.system.enabled)
+            .map((e) => new TokenEffect(e));
+
+        return [...super.appliedEffects, ...fromEffects];
+    }
+
     // Temporary effects are displayed on the token, so hijack it and include effects
     get temporaryEffects() {
```
ian612, line 52: Does this still need to be here in v14/do we still need to override this as well?

src/module/token/token-effect.js
```diff
@@ -16,6 +16,10 @@ export class TokenEffect {
         return this.#effect.parent;
     }
 
+    get showIcon() {
```
ian612, line 25: Should this return 1/0 or boolean `true`/`false`?

ian612 (approved):

src/module/actor/actor.js
danimrath: Added comment, the code is still present in V14 they just aren't calling it. We should keep the override here just in case they decide to call it again.

src/module/token/token-effect.js
danimrath: Added explanation of the values. Const values from Foundry can be 0, 1 or 2.

PR #1773, opened 2026-08-22
Combat Time Changes

danimrath:
Currently if a player without higher permissions ends their turn it throws an error and also the time isn't moving properly.
This change makes it so only the GM's session calls advance time and makes several functions/calls async await.

image


ian612 (commented):
Don't see any issues, looks good to me

ian612 (approved):

ian612 (approved):

